PR #47, opened 2013-05-12
Enhanced EEPROM storage, fixes to pen up/down logic

bytemedwb:
Primarily fixes to store additional data in EEPROM and make sure data is written
after an update in the GUI.

Fixed Pen up/down logic in Arduino code where Up angle is less than Down angle.

Fixed Pen up/down logic in Pause code.


i-make-robots:
These fixes have been incorporated into the latest version, sans merge.  You spotted it first so I'll get you into the credits regardless.

PR #60, opened 2015-02-10
Allowed for serial connections on Linux

chrsmlls333:
Removed default to Windows so that Linux is also usable. Allows for successful printing on Raspberry Pi among other systems. 
(Tested on Windows 8.1 and Raspbian)

Code referenced: http://www.mkyong.com/java/how-to-detect-os-in-java-systemgetpropertyosname/

screenshot 11


i-make-robots:
This is sweet.  Can you detail your Rpi setup, please?  I will include it in the setup documentation.

chrsmlls333:
Quick Guide here. May want to include the bash script described along with the batch ones in the next release.
https://www.marginallyclever.com/forum/viewtopic.php?f=19&t=4947

PR #220, opened 2017-07-14
Say hi to robot when establishing a new connection

drewler:
Send M100 to robot so it can identify itself.

### DIFF
--- a/java/src/main/java/com/marginallyclever/communications/serial/SerialConnection.java
+++ b/java/src/main/java/com/marginallyclever/communications/serial/SerialConnection.java
@@ -85,7 +85,6 @@ public final class SerialConnection implements SerialPortEventListener, NetworkC
 		connectionName = portName;
 		portOpened = true;
 		waitingForCue = true;
-		this.sendMessage("M100\n");
 
 	}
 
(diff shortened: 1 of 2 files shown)


i-make-robots:
Makelangelo robots already identify themselves when they connect.  Makelangelo-firmware calls help() on setup().  help() is also called on M100.  What does this fix?

drewler:
I'm sorry, this is my first pull request and I didn't provide much information.

With the board I'm using, the "Hello message" wasn't reaching the Java application. Also, because it is only sent from the setup() function, which only runs at startup, I couldn't reconnect.

I fixed this by sending an M100 command to the robot whenever a new serial conection is created from the Java application, which triggers the help() function at firmware level. This way, both the first connection and all future reconnections work. I though the fix could be useful and because of that, I submitted it.

i-make-robots:
Aha... Well... the fix seems to be in a class that deals exclusively with communications to a generic serial device.  At that level the classes should have no awareness of gcode.  It would probably be better placed in MakelangeloRobot.  That way if a new transport layer is written (say, TCP/IP) then the M100 will still be sent because it happens outside of the serial connection creation.  Does that sound ok?

drewler:
Yes, you are right. I'll try to come up with a generalized solution in MakelangeloRobot and update this pull request then. Thanks!

drewler:
I moved the action of sending the message to the method `openConnection(NetworkConnection c)` of `MakelangeloRobot`. What do you think?

i-make-robots:
Sure, that'll probably work.  I wonder what happens when a new machine connects and the software gets UID -1 twice...  hmm...

drewler:
Do you mean when it has to get an UID assigned from the server?

PR #256, opened 2019-10-28
Here are 2 issues I found when trying to use Scratch2. (haven't tested Scratch3)

mishafarms:
Don't assume that child 0 is the child we want. Find the first childwith a scripts field

Don't assume that there are global variables.


mishafarms:
These are in the newest pull request

i-make-robots:
Scratch3 is very hard to process and I haven't finished the code.  I don't see an obvious improvement in the features.  It's probably only going to help in a school where they can't install software but they CAN run a website and export.

PR #270, opened 2020-06-06
Reset zoom with keyboard shortcut.

Jip-Hop:
Like in Google Chrome, reset zoom with CMD+0.

### DIFF
--- a/src/main/com/marginallyclever/makelangelo/Makelangelo.java
+++ b/src/main/com/marginallyclever/makelangelo/Makelangelo.java
@@ -254,7 +254,9 @@ public final class Makelangelo
 				KeyStroke.getKeyStroke(KeyEvent.VK_EQUALS, Toolkit.getDefaultToolkit().getMenuShortcutKeyMask()));
 		menu.add(buttonZoomIn);
 
-		buttonZoomToFit = new JMenuItem(Translator.get("ZoomFit"));
+		buttonZoomToFit = new JMenuItem(Translator.get("ZoomFit"), KeyEvent.VK_0);
+		buttonZoomToFit.setAccelerator(
+				KeyStroke.getKeyStroke(KeyEvent.VK_0, Toolkit.getDefaultToolkit().getMenuShortcutKeyMask()));
 		buttonZoomToFit.addActionListener(this);
 		menu.add(buttonZoomToFit);
 


i-make-robots:
You're starting to cover something I have kind of fixed in robot overlord - everything that is clickable should be an undoable action with a shortcut that can be tweaked by the user.  But that's a tomorrow goal!  I love that you're making these improvements.  You rock!  What's your IG handle?  I wanna follow you :)

Jip-Hop:
Haha you rock! I don't really have social media. I've got a WordPress website though! http://jipdebeer.media <http://jipdebeer.media/>
What is Robot Overload? Does it work with Makelangelo machines?

By the way the reason I made this small change is that scrolling to zoom with the trackpad on MacBooks doesn't work great.
So when I forget about this I can now reset the zoom and just use the keyboard shortcuts to control zoom.

Not a real fix, just a workaround :)

i-make-robots:
Fair.  I have a 10+ year old macbook pro, i don't use it for makelangelo much any more.  when i did, I don't think I ever zoomed away from the main view much except to command+/-.

Will you post your drawings there?  I'm so excited right now, I just wanna barf all my thoughts about where I've been trying to go with this code.  The more I work it the more I think there's really two separate apps here - one to generate drawings and one that just optimizes drawings and runs the robot.  That's why the dev branch has this "art pipeline" that is... definitely first-draft quality.

Jip-Hop:
Please do barf ahead! xD I'm curious where you're taking it. I think I'll start experimenting with Makelangelo app first, but later I hope to do some real-time drawing by directly sending gcode over serial. Don't know exactly what I'll draw, probably visualising sensor data of some sorts. Will probably look messy. For now I'm still building the machine. It's going to be approximately 210cm heigh and 236cm wide, using Arduino + ramps 1.4 + HC6 bluetooth + endstops + bearing tensioners. I've printed the pen holder, it's really nice. Still waiting for some extension cables to be delivered, and the weights too. Oh, and I've got a 1x2m whiteboard 'sticker' which will be my test canvas for when I don't want to waste paper.

i-make-robots:
Nice.  The longer I work with the conversion styles the more I see repetition of ideas happening.  So it would be neat to evolve towards a node-based programming where effects can be combined without rewriting the code every time.  Another is that I'd like to generate super huge drawings and then split them into N*M paper-sized images.  What else... none of the jpanels are testable on their own - hell, many classes are interlinked so unit testing is basically impossible.  there's no way to simulate a drawing in the app or show n% of a drawing (like 3d slicers can show a print after N layers).  These are all wants that I haven't got to yet.

i-make-robots:
Another way would be to make a processing library that generates a text file Makelangelo software can read and then lose all generators & converters from inside the app, cut the app way way back to "fit image on paper, simulate, draw."

Jip-Hop:
Cool ideas! It's quite convenient that there are image conversion options right inside the app, but I see benefits of a minimalized app too. Would be much easier to maintain and keep bug free :) Actually I was planning on opening a couple of issues if you don't mind. Could be user error, but maybe they're real bugs. I'll be trying out the dev branch by the way, because I connect with Bluetooth.

i-make-robots:
Please do.

PR #272, opened 2020-06-07
Fix javadoc error.

Jip-Hop:
See https://bugs.openjdk.java.net/browse/JDK-8212233.
Now it builds without errors for me too.


i-make-robots:
Nice!

PR #277, opened 2020-06-11
Update JogAmp (and changed hardware version...)

Jip-Hop:
Whoops I tried to make two separate pull requests, one for the Makelangelo 3.3 hardware version number, and a second one for updating JogAmp. But now both changes are included in this pull request... sorry.

What I did to update JogAmp:

Change `pom.xml` to reflect current version number of JogAmp.

Download [latest recommended JogAmp](https://jogamp.org/deployment/jogamp-next/fat/).

Import `jogamp-fat.jar` like so:

Import -> Maven -> Install or deploy an artifact to a Maven repository.
Then moved the jogamp directory from /path/to/user/home/directory/.m2/repository/org/ to the org directory in this projects local Maven repository (java/local-maven-repo/org).

With this version Makelanglo [no longer silently crashes at startup on Windows 10](https://github.com/MarginallyClever/Makelangelo-software/issues/271). It runs with JRE, no need for JDK. Unfortunately it doesn't solve [the MacOS drawing glitch](https://github.com/MarginallyClever/Makelangelo-software/issues/264#issuecomment-642037278).


i-make-robots:
Does this mean the wiki getting started page needs jogamp-fat instructions?

Jip-Hop:
I don't think so. Jogamp-fat is included in the repository. Cloning the repo should include it and building should work with the current instructions.

Or what do you mean?

PR #285, opened 2020-06-25
Reduce CPU usage.

Jip-Hop:
Makelangelo takes up 160% CPU usage on my MacBook Pro, when it's basically idle. The `FPSAnimator` which repaints the `drawPanel` at 20fps is wasting lots of resources. Most of the time nothing has changed and the `drawPanel` doesn't need to be repainted.

I now reduced the framerate to 1 repaint per second, so the interface will continue to be updated (albeit with slight lag). But I've added explicit calls to `repaint()` when certain events occur, like dragging the viewport, or zooming. That way, these events will repaint immediately, there won't be any lag, and we won't be constantly repainting when nothing changed.

Ideally the `FPSAnimator` would be removed completely, and all events (like the manual driving commands) would call `repaint()`.

With the changes in this pull request the CPU usage dropped to 15%. Now I can keep Makelangelo open more comfortably.


i-make-robots:
Commits to master should only happen when a new version is released or a hotfix for a bug in a release.  If you could make this PR to the main dev branch instead, I'm happy to make it happen.

Jip-Hop:
Whoops that's what I was trying to do. But a popup on GitHub tricked me.

PR #306, opened 2020-11-15
added fields to shift the paper on the machine. it no longs needs to …

gsohler:
paper does not need to be centered anymore in the plotter

Fixed Code to import SVG to use actual Paper size. old code only worked correctly when it was 0/0


i-make-robots:
Does the paper shift work with loading other file formats, or is it only applicable to svg?

gsohler:
Hi Dan,
Paper shifts works already for import-svg and for Gosper fractals
,, under normal circumstances when Art Pipeline is set to leave the sizes.(else) the shift is cancelled.
So i think its the better idea to implement the shift in the art pipeline as it will affect all generators.
Actually i suggest to cancel this request ;)

Thank you for using the package box though ...

i-make-robots:
I agree, I didn't know how to say politely that art pipeline was the better location.

PR #307, opened 2020-11-22
added generator for big parcel boxes

gsohler:
Added a generator to create custom pacakge boxes for shipping. usually you cannot print them with your desktop printer,
because they are way too latrge.
Makelangelo is perfect for that.


i-make-robots (approved):

i-make-robots:
Nice.

PR #308, opened 2020-11-24
Add paper shifts

gsohler:
Finally implemented Paper shifts through the art pipeline.
now applies to all generators.

### DIFF
--- a/src/main/com/marginallyclever/artPipeline/ArtPipeline.java
+++ b/src/main/com/marginallyclever/artPipeline/ArtPipeline.java
@@ -413,37 +413,6 @@ public class ArtPipeline {
 
 	
 	
-	/**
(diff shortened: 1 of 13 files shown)


i-make-robots:
Nice.  hey, if you want there is a Discord channel where some firmware fans are meeting: https://discord.gg/Vekf7bsh

i-make-robots:
I see crop now happens before scale/fit.  does that leave some lines outside the margins?

Later I see that paper size does not (immdiately, obviously) use margin any more.  Are paper margins gone?

gsohler:
It was intuitive to me to have crop after scaling but you dont need to keep that order.
Paper Margins are not gone. They are implemented  now reducing the scale.
You are free to pull in parts only.

PR #310, opened 2020-11-29
Polyeder

gsohler:
This is the Polyeder generator i was talking about.
Its basically just some basic trigonometry.
I am wondering if you can decode the integer codes in the main function, the code  can.


i-make-robots:
> I am wondering if you can decode the integer codes in the main function, the code can.

Please elaborate, I don't understand "decode the integer codes"

gsohler:
In the Code there are huge integer arrays like

int [] {4,4,0,0,1,1, ...

i am wondering if you understand the meaning....

i-make-robots:
It would help if the code had more comments to explain what it is doing.  I infer that '1' means draw a square and 3-8 are N-gons.

gsohler:
The code works recursively interpreting the number sequence from the list. So 4 draws a 4-gon(square) and afterwards recursively descends drawing something to all its edges. '1'  instead just draws a flap, and '0' does nothing.

i-make-robots:
Yes, and I see the recursion.  Maybe a comment block at the top of the class that has your name, @since version, and a description of how it works.

Are you aware that Turtle can rotate with turn(degrees) and make relative moves forward(stepSize)?

i-make-robots:
I'm pretty sure the rhombicosidodecahedron has too many flaps.  there are flaps against other flaps?  Maybe that's intentional.

Anyways, nice work!  I added comment to describe the system.

i-make-robots:
https://twitter.com/aggrav8d/status/1333513002922823682?s=20

PR #311, opened 2020-11-29
Customplotter

gsohler:
Everytime when i actually want to plot something in my plotter, i need to make sure, that the belt len and home is correctly set but of course you are not interested in that as its not working with Makelangelo5

So i have now made a conditional , so this applies only for Custom Plotters.

I would really appreciate have this pr in place.
Please check the changes and raise any concerns.

### DIFF
--- a/src/main/com/marginallyclever/makelangeloRobot/settings/MakelangeloRobotSettings.java
+++ b/src/main/com/marginallyclever/makelangeloRobot/settings/MakelangeloRobotSettings.java
@@ -188,10 +188,7 @@ public final class MakelangeloRobotSettings implements Serializable {
 	}
 
 	public Point2D getHome() {
-		if(hardwareVersion.equals("0")) {
-			return new Point2D(0,limitBottom);
-		}
-		return getHardwareProperties().getHome();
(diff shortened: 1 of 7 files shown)


i-make-robots:
I have a business with employees.  I barely get to run the M5.  There's no open ticket describing the problem in the M3.  So Yes, I didn't know about it and No, I *do* care.  I take offense at that because I work hard to keep everyone happy.

Anyways.

getHardwareProperties().getHome() implies custom stuff should be in MakelangeloCustomProperties::getHome().  What say you?

i-make-robots:
The intent was that everything to manage a given hardware version would be in the appropriate hardware class.  If settings are being reset - but only on custom machines - then something fishy is going on.  ideally settings are applied once to each machine and never changed again.

gsohler:
Sorry, did not want to offend you.
Of course i did not want my pr to change ANY  plotters other than than the Custom Plotter(including M3 and M5)
My Issue does not relate to M3 but to Custom Makelangelo.

HardwareProperties does not have an abstract function for getGCodeConfig(),
i will suggest one.

gsohler:
will close this request and put the corrected version

PR #314, opened 2020-12-04
Simplify

gsohler:
This is my implementation of simplify called simplify1. Its a mixture of your simplify and reorder and not activated in this PR.
It use Quadtree's to store points, so its very efficient storing lots of points. (believe its O(N*log(N) )
This implementation  knows which line segment ends sit at the same point but it does not know which line ends are near by
as it would require a O(N^2) operation.
It can simplify duplicate parallel and antiparallal lines, but not partially overlapped lines with neither point coincident .
Finally it also reorders the points while requesting each pen change only once.
It has comparable results to yours but i believe its more efficient for huge amounts of data.

### DIFF
--- a/src/main/com/marginallyclever/artPipeline/ArtPipeline.java
+++ b/src/main/com/marginallyclever/artPipeline/ArtPipeline.java
@@ -475,7 +475,7 @@ public class ArtPipeline {
 
 	public void simplify1(Turtle turtle, MakelangeloRobotSettings settings) {
 		int os, ns;
-		int i, j, k;
+		int i, j;
 		double x,y;
 		PointBranch ind = null, oldind = null;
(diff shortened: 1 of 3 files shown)


i-make-robots:
Why can't I see pr 314 in my github app? :(

i-make-robots:
Looks good.  needs more comments to explain what is going on.

gsohler:
i can see pr314

[image: image.png]

i-make-robots:
Hello,

I didn't merge this earlier and I'm sorry.  There is a new SimplifyTurtle system.  Could you make SimplifyTurtle2 so we can run them side by side?  I would be more eager to merge such a change.  Your work deserves to be included.

PR #320, opened 2021-01-18
Dev

Georgwei:
added Chinese（Simplified） translation


i-make-robots:
Did you find any places where the translation is challenging because the text is forced into a funny order?  
For example, where it english might say "A [number] B" and chinese might be better with "A B [number]"?

Georgwei:
I believe there is no problem with the translation itself.

Georgwei:
Are you referring to the meaning of Chinese translation or the xml file
format?
I believe the meaning of Chinese translation is not too problematic.

Dan Royer <notifications@github.com> 于2021年1月19日周二 上午2:22写道：

PR #355, opened 2021-11-01
Fix unit tests and extract GUI Tests

coliss86:
- Removed the dependency `com.googlecode.json-simple:json-simple` as `org.json:json` is already used instead
- 2 Units Tests where failling because they started the GUI of the application and there is no human interaction possible in Github Action. Thus, the test was killed after 6 hours (maximum duration of a Github action step). I have added a configuration for the maven plugin `maven-surefire-plugin` to skip file whose class name ends with `GUITest` and `IntegrationTest`. Usually, tests with a graphical interface are not runned in the CI and are not called "Unit Tests" but "Integration Tests". In order to skip those tests in the CI, I have renamed : 
  - `src/test/com/marginallyclever/makelangelo/MakelangeloTests.java` to `src/test/com/marginallyclever/makelangelo/MakelangeloGUITest.java`
  - `src/test/com/marginallyclever/makelangeloRobot/MakelangeloRobotPanelTest.java` to `src/test/com/marginallyclever/makelangeloRobot/MakelangeloRobotPanelGUITest.java`
    You can still run those tests with your IDE
- I have modified some unit tests to use the `Assertions` API : `assertEquals()`, `assertNotNull()`. Those methods should be used instead of the method `assert()` as the last one is part of the JDK and is only enable when launching the JVM with the `-ea` option. It is rarely used. In unit test, the `assertEquals()` produce meaning full message when the value is not expected.
- I have split the building process in 2 steps in order to find easily test results in logs : "Build and Test with Maven" and "Package with Maven"

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -25,6 +25,7 @@ jobs:
 
     # from https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
     - name: Build with Maven
+      timeout-minutes: 15
       run: ./mvnw -B package --file pom.xml
 
     - name: prepare package for release
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -24,11 +24,15 @@ jobs:
           ${{ runner.os }}-maven-
 
     # from https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
-    - name: Build with Maven
+    - name: Build and Test with Maven
       timeout-minutes: 15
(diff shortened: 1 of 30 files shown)


i-make-robots:
Seems good so far...

i-make-robots:
I suck at package management and I tried to get CI working.  The fact you're making it work?  I have deep admiration for you right now.  Thank you!!

coliss86:
For the moment, I was just looking for the tests whose where failing the CI by disabling them. Now it's time to make something more permanent :)

PR #356, opened 2021-11-08
Update doc

coliss86:
Hi,

Here is an update of the README with a screenshot of the new interface as well as instructions to get a pre-compiled version from the nightly build.

Regards


i-make-robots:
“Everybody wants to save the Earth; nobody wants to help Mom do the dishes.”
― P.J. O'Rourke

You, sir, are a champion of all that is good in the world.

PR #361, opened 2021-11-18
Completely re-translated english.xml into german

captFuture:


i-make-robots:
Nice!  I and every german speaking user thanks you.

i-make-robots:
if you bump the version number it will appear in the nightly build correctly numbered.

PR #364, opened 2021-11-26
Add @DisabledIfEnvironmentVariable to skip GUI unit test in the CI

coliss86:
Hi, 

Here a new set of improvement for tests which starts a graphical interface during their execution:
- Add a new annotation [`@DisabledIfEnvironmentVariable`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/condition/DisabledIfEnvironmentVariable.html#named()) for this kind of test which disable their execution when run in github CI.
- `-Djava.awt.headless=true` for maven while make tests fails as soon as they try to open a window => you'll find them easily and thus you'll be able to add `@DisabledIfEnvironmentVariable`
- Revert the renamming of few Tests from `*GUITest` to `*Test` that I have done in https://github.com/MarginallyClever/Makelangelo-software/pull/355 which is not needed anymore

Tell me what you think


i-make-robots:
Why is this better than using <exclude>?

coliss86:
With this, you can still run `./mvnw verify` on your local computer to run all the test in one shot even those with the GUI. Previously, those tests where not run anymore with maven and you had to launch them with Eclipse or Intellij explicitly.

Tests with `@DisabledIfEnvironmentVariable` are excluded in Github CI due to the fact that the environment variable `CI` is automatically added by Github.

coliss86:
Now, when tests need a graphical env, they will fail with the following error: 
```
java.awt.HeadlessException
	at com.marginallyclever.makelangelo@7.30.1/com.marginallyclever.makelangelo.TranslatorTests.startTranslatorTwiceTest(TranslatorTests.java:34)
```
see here for an exemple: https://github.com/coliss86/Makelangelo-software/runs/4328561429?check_suite_focus=true#step:5:3512

PR #366, opened 2021-12-17
Fix FillPage (infinit loop -> GUI Freeze) and GosperCurve (Travels to Draws)

PPAC37:
Avoid an infinit loop if penDiameter <=0


i-make-robots:
does turtle need to move to start of gosper curve before pen down?

PPAC37:
From what I understand the gosper curve will not be drawn if the pen is not down and the curve begins immediately.
Or am I missing something?
https://github.com/MarginallyClever/Makelangelo-software/blob/74ecc1353899d127748074923ffdddabc282aa5c/src/main/com/marginallyclever/makelangelo/makeArt/turtleGenerator/Generator_GosperCurve.java#L36-L40

i-make-robots:
I mean to say that if you pen down before moving to the start of the drawing the machine will home; pen down; draw a line to the start; and then proceed as normal.  that line from home to start is probably undesirable.

PPAC37:
In the generated g-code i get the pen down on the good position. (not on the Home position)
But i have also make an adjustment in the g-code generator ... I added a pen Up just after the Home (G28) ... 

it might be necessary to see, as with the 3D printer slicers, to have a starting g-code and an end g-code that can be modified by the user ?)

PPAC37:
My excuses.
I realize that I have messed up ... there are some commits that should not be here ... I will try to undo them.

PR #370, opened 2021-12-19
Remove PreferencesHelper & Translator in MakelangeloTest as they are not needed

coliss86:
Hi,

here a small modification to enable MakelangeloTest to be run in the CI: PreferencesHelper and Translator are not needed ti read the version of the software

Regards


i-make-robots:
Well spotted!

Apparently the defaultLookAndFeel aren't needed, either?

PR #376, opened 2021-12-20
Exit when all elements are stopped

coliss86:
when clicking on exit in the "file" menu, the JVM does not exit. Here is a fix.


i-make-robots:
I'm pretty sure the JFrame closing is supposed to kill the thread and so system.exit() is not supposed to be called.

i-make-robots:
on my system `File > Quit` works fine, no lingering thread.

PPAC37:
For me, the JFrame close on File -> Quit. 
But if i run the jar via a command line, i dont get back the prompt ... the JVM is still running ... 
```
6@q6-pc:~/Downloads$ java -jar makelangelo-nightly-20211220-a83707.jar
log dir=/home/q6/Downloads/

```
i need to do do ctrl+c get back the promt ...
```
q6@q6-pc:~/Downloads$ java -jar makelangelo-nightly-20211220-a83707.jar
log dir=/home/q6/Downloads/
^Cq6@q6-pc:~/Downloads$ 
```

coliss86:
hum weird :-(
When launching from a console or with IntelliJ, it does not exit, the window is still closed.

i-make-robots:
Then something more serious is happening. The h frame is not drawing but has not completely closed?

coliss86:
yes, nothing is visible

coliss86:
do you want to investigate more or as the application is closing, this call is enough ?

coliss86:
I get the same behaviour when launching `ScaleTurtlePanel.main()`: the window close but the jvm remains up...

i-make-robots:
mainFrame.setDefaultCloseOperation(JFrame.DO_NOTHING_ON_CLOSE); is the default behavior.
mainFrame.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE); is set during mainFrame.windowClosing() --> onClose().
mainFrame.windowClosed() comes after that and should correctly exit.

FWIW, there's also https://tips4java.wordpress.com/2009/05/01/closing-an-application/ which makes a nice ExitAction.  still doesn't explain why the process fails to quit for some people.

PR #377, opened 2021-12-20
Replace System.out.println with Log.message

coliss86:


i-make-robots:
Following your good example I did the same to all Robot Overlord calls.  Thanks!

PR #379, opened 2021-12-21
Move the save to gcode button in the menu Makelangelo

coliss86:
Hi,

I moved the "save gcode" button in the Makelangelo menu for direct access: 
image

Here is the updated plotter interface:
image

### DIFF
--- a/src/main/com/marginallyclever/makelangelo/plotter/plotterControls/SaveGCode.java
+++ b/src/main/com/marginallyclever/makelangelo/plotter/plotterControls/SaveGCode.java
@@ -1,14 +1,10 @@
 package com.marginallyclever.makelangelo.plotter.plotterControls;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.OutputStream;
-import java.io.OutputStreamWriter;
+import java.io.*;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 
-import javax.swing.JComponent;
-import javax.swing.JFileChooser;
+import javax.swing.*;
 import javax.swing.filechooser.FileNameExtensionFilter;
 
 import org.apache.commons.io.FilenameUtils;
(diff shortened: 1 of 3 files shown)


PPAC37:
I Like this ! :)

Maybe adding a Mnemonic and an Accelerator on the JMenuItem bSaveToSD ?
something like :
```
bSaveToSD.setMnemonic(KeyEvent.VK_S);
bSaveToSD.setAccelerator(KeyStroke.getKeyStroke("alt S")); // or "ctrl S" ?
```
But this can create the need to user modification of that (different language ?= different Mnemonic / Accelerator) ... so maybe not. This is to discut.

Another discussion is why not just "duplicate" / "recreate" rather than move (remove from panel)?

coliss86:
When saving on the SD card, the panel is not necessary, so you don't have to open it anymore

src/main/com/marginallyclever/makelangelo/plotter/plotterControls/SaveGCode.java
```diff
@@ -19,27 +15,26 @@
 import com.marginallyclever.makelangelo.turtle.TurtleMove;
 
 public class SaveGCode {
-	private JFileChooser fc = new JFileChooser();
-	private FileNameExtensionFilter filter = new FileNameExtensionFilter("GCode", "gcode");
-	
+	private final JFileChooser fc = new JFileChooser();
+
 	public SaveGCode() {
+		FileNameExtensionFilter filter = new FileNameExtensionFilter("GCode", "gcode");
```
i-make-robots, line 21: why is this recreated on every save?  to reduce memory?

src/main/com/marginallyclever/makelangelo/plotter/plotterControls/SaveGCode.java
```diff
@@ -51,52 +46,48 @@ private String addExtension(String name, String [] extensions) {
 		return name + "." + extensions[0];
 	}
 	
-	private void save(OutputStream outputStream,Turtle turtle,Plotter robot) throws Exception {
+	private void save(String filename, Turtle turtle, Plotter robot) throws Exception {
 		Log.message("saving...");
 		
-		OutputStreamWriter out = new OutputStreamWriter(outputStream);
+		try (Writer out = new OutputStreamWriter(new FileOutputStream(filename))) {
```
i-make-robots, line 52: why is there a try here if there is no catch?  Why is it better to create the OutputStreamWriter here than where it was?

I love when format-all makes everything angry red....

src/main/com/marginallyclever/makelangelo/plotter/plotterControls/SaveGCode.java
coliss86: it is a small object only used in this method. it was declared as an instance field of the class and only used some times, no memory optimization but I like short live object.

src/main/com/marginallyclever/makelangelo/plotter/plotterControls/SaveGCode.java
coliss86: in the original implementation, the `close()` was not called if an exception was thrown is the method. Here, it is the [try with resources statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) which will close any stream open between the brace at the beginning.
I didn't add a catch because the method already throws Exception which tells that the parent is handling the exception for me.

i-make-robots:
I'm going to merge this.

It will probably be an issue later when support is re-introduced for different sizes and styles of plotter.

PR #381, opened 2021-12-21
Set the "Scale window" larger and select the lock by default

coliss86:
I increased the size of the scale window in order to see better the values.
The "lock 🔒" is selected by default, and the button up and down increment by 1 unit instead of 0.1.

Before:
image

After:
image

### DIFF
--- a/src/main/com/marginallyclever/makelangelo/ScaleTurtlePanel.java
+++ b/src/main/com/marginallyclever/makelangelo/ScaleTurtlePanel.java
@@ -1,26 +1,14 @@
 package com.marginallyclever.makelangelo;
 
-import javax.swing.JButton;
-import javax.swing.JCheckBox;
-import javax.swing.JComboBox;
-import javax.swing.JComponent;
-import javax.swing.JDialog;
-import javax.swing.JFrame;
-import javax.swing.JLabel;
-import javax.swing.JOptionPane;
-import javax.swing.JPanel;
-import javax.swing.JSpinner;
-import javax.swing.SpinnerNumberModel;
+import javax.swing.*;
 import javax.swing.event.ChangeEvent;
 
 import org.apache.batik.ext.swing.GridBagConstants;
 
-import java.awt.Dimension;
-import java.awt.GridBagConstraints;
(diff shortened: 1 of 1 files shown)


src/main/com/marginallyclever/makelangelo/ScaleTurtlePanel.java
```diff
@@ -31,16 +19,16 @@
 	/**
 	 * 
 	 */
+	@Serial
```
i-make-robots, line 22: what is this for?

src/main/com/marginallyclever/makelangelo/ScaleTurtlePanel.java
```diff
@@ -226,12 +217,9 @@ public static void main(String[] args) {
...
 
-		try {
-			JFrame frame = new JFrame(ScaleTurtlePanel.class.getSimpleName());
-			frame.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE);
-			runAsDialog(frame,new Turtle());
-		} catch (Exception e) {
-			JOptionPane.showMessageDialog(null,e.getMessage(),"Error",JOptionPane.ERROR_MESSAGE);
-		}
+		JFrame frame = new JFrame(ScaleTurtlePanel.class.getSimpleName());
+		frame.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
```
i-make-robots, line 221: EXIT_ON_CLOSE causes system.exit() to be called, as discussed elsewhere.  DISPOSE_ON_CLOSE might be what you're looking for.

i-make-robots (commented):
I wasn't trying to start a review.  but my comments stand.

src/main/com/marginallyclever/makelangelo/ScaleTurtlePanel.java
coliss86: the purpose of this modification is to remove the `catch` and let the exception be handled by the `main()` method which will display it in the console and exit in error if something wrong happen. Nothing related to `EXIT_ON_CLOSE` or else ;-)

PR #382, opened 2021-12-21
Do not show the "load file" window when loading a recent file

coliss86:
The dialog box (see below) to load a file is not shown anymore when a recent file is loaded.
The file is directly loaded without any further click.

image

### DIFF
--- a/src/main/com/marginallyclever/makelangelo/Makelangelo.java
+++ b/src/main/com/marginallyclever/makelangelo/Makelangelo.java
@@ -420,24 +420,23 @@ public final class Makelangelo {
 			LoadFilePanel loader = new LoadFilePanel(myPaper,filename);
 			loader.addActionListener((e)-> setTurtle((Turtle)(e).getSource()) );
 			previewPanel.addListener(loader);
-			if(filename!=null && !filename.trim().isEmpty() ) {
+			if (filename!=null && !filename.trim().isEmpty()) {
 				loader.load(filename);
+			} else {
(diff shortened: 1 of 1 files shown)


i-make-robots:
What happens if your recently loaded file was an image and you are trying to change the conversion options?

coliss86:
good question, I never see a conversion option, may be it is related to #383

i-make-robots:
I believe #383 is fixed now, so you can try it with images and vectors to see the difference.

coliss86:
Ok I understand better the original behavior. I then propose another implementation: when a recent SVG is loaded, no dialog box, otherwise it is shown.

WDYT ?

i-make-robots:
One of those "you couldn't know" things... inside loading SVG there are some options that would be nice to make visible, like conversion quality of curves.  I never got around to it.  As a rule I don't cut off an option that could be expanded on later.  Maybe that's a dumb thing to do, but it hasn't hurt...?

coliss86:
My bad, but I do not see any option: 

image

Am I missing something ?

i-make-robots:
I wasn't clear.  Today there are no sub-options for SVG.  In the future it would be nice.  Instead of making a special exception for only SVG files in the parent class - which is hard to remember and prone to breaking - I leave a blank window and room to expand later.  it's one click and no one is hurt.  if they want they can click the open button and pick another svg.  I'd rather not change the workflow if I can avoid it.

coliss86:
Ok, I understand, it makes sense. I'm closing this PR.

PR #387, opened 2021-12-22
Add missing translation

### LINKED ISSUE #386: Translations missing
**Steps to make the bug happen**
1. Import a new picture
2. See error
image

**What was supposed to happen**
I think that some translations are missing

**Platform (please complete the following information):**
 - Version: master 8fe5def5b94775d3ba8b3

coliss86:
Fix #386


i-make-robots:
Please remove the sandy noble parts.
Please add the class name in front of each string that is used only by that class, as I did in sandy.  Eventually we'll get these all cleaned up.

PR #394, opened 2021-12-22
Makelangelo JFrame KEY_WINDOW_Y value to "windowY" // was "windowX"

PPAC37:
unless it was done on purpose not to position the window at the right coordinates for the next launch. ;)


i-make-robots:
Good catch!

PR #395, opened 2021-12-23
Fix LoadFilePanel panel frozzen after Open Cancelled.

PPAC37:
It seems to me that this permutation must be made. (unless it's normal ... or on purpose.)

Otherwise when you open a recent file and if you use the "Open" button which opens a JFileChooser that you "Cancel", the controls on the panel are no longer usable. 
(and as the image has not changed ... it's disagreeable to have to close the LoadFile dialog to reopen it ...)

### DIFF
--- a/src/main/com/marginallyclever/makelangelo/makeArt/io/LoadFilePanel.java
+++ b/src/main/com/marginallyclever/makelangelo/makeArt/io/LoadFilePanel.java
@@ -65,8 +65,9 @@ public class LoadFilePanel extends JPanel implements PreviewListener {
 	}
 	
 	public void chooseFile() {
-		mySubPanel.removeAll();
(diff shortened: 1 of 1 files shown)


i-make-robots:
- click File > Open (or) File > Open Recent > (pick anything).  LoadFilePanel appears.  Main menu becomes inactive.
- Click Open File.  Open dialog appears.
- Cancel or 'x' the dialog.  Open dialog disappeasr.  LoadFilePanel is still there.
- Close LoadFilePanel.  Main window is active, main menu becomes active.

Where is the problem?

PPAC37:
Sorry, I worded badly.
With pictures;):

from there : 
image

then if you do : "Open file ..." and "Cancel" the JFileChooser.

image

when you return to the "LoadFilePanel" the subpanel is not usable / frozzen :

image

i-make-robots:
Good catch!

PR #397, opened 2021-12-23
Fix Language name value from preferences no match.

PPAC37:
To ask the user to select a valid language name if the preference language name does not match a language name value in the language file.

To avoid some null issues in Translator.get(String key)
or else you get this kind of problems :

```
Exception in thread "AWT-EventQueue-0" java.lang.ExceptionInInitializerError
...
Caused by: java.lang.NullPointerException: Cannot invoke "com.marginallyclever.makelangelo.TranslatorLanguage.get(String)" because the return value of "java.util.Map.get(Object)" is null
	at com.marginallyclever.makelangelo.Translator.get(Translator.java:223)
	at com.marginallyclever.makelangelo.makeArt.io.vector.LoadScratch3.<init>(LoadScratch3.java:68)
	at com.marginallyclever.makelangelo.makeArt.io.vector.TurtleFactory.<clinit>(TurtleFactory.java:13)

```

### DIFF
--- a/src/main/com/marginallyclever/makelangelo/Translator.java
+++ b/src/main/com/marginallyclever/makelangelo/Translator.java
@@ -66,7 +66,19 @@ public final class Translator {
 	static public boolean isThisTheFirstTimeLoadingLanguageFiles() {
 		// Did the language file disappear?  Offer the language dialog.
 		try {
-			if (doesLanguagePreferenceExist()) {
+			if (doesLanguagePreferenceExist()) {  
+
+                        // Does the language preference have a language name value 
+                        // that matches a language name value in an available language .xml file
+                        String languageNameFromPref = languagePreferenceNode.get(LANGUAGE_KEY, defaultLanguage);
+                        if ( ! languages.keySet().contains(languageNameFromPref)){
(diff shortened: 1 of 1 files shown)


src/main/com/marginallyclever/makelangelo/Translator.java
```diff
@@ -66,7 +66,19 @@ static public void start() {
 	static public boolean isThisTheFirstTimeLoadingLanguageFiles() {
 		// Did the language file disappear?  Offer the language dialog.
 		try {
-			if (doesLanguagePreferenceExist()) {
+			if (doesLanguagePreferenceExist()) {  
+
+                        // Does the language preference have a language name value
```
coliss86, line 71: the formatting does not match the rest of the file: you have added 25 space char instead of 4 tabs.

i-make-robots:
nice one!

src/main/com/marginallyclever/makelangelo/Translator.java
PPAC37: > the formatting does not match the rest of the file: you have added 25 space char instead of 4 tabs.

Sorry for the spaces instead of the tabs (bad configuration of my editor ...) ... by the time I change, the request has been merged.
Should I redo a PR?

PR #400, opened 2021-12-23
javadoc build errors fix.

PPAC37:
No review just so javadoc can be build.


src/main/java/com/marginallyclever/makelangelo/makeArt/imageConverter/ImageConverter.java
```diff
@@ -16,7 +16,7 @@
 import com.marginallyclever.makelangelo.turtle.Turtle;
 
 /**
- * Converts a {@link BufferedImage} to {@link Turtle}
+ * Converts a {@link java.awt.image.BufferedImage} to {@link Turtle}
```
i-make-robots, line 19: TransformedImage, not BufferedImage

PR #403, opened 2021-12-24
Paper pref

PPAC37:
Seems to do the job of saving and reloading paper settings preferences ...

But it's not clean coding because I didn't do a needs analysis, I just tried to make it work.

And moving the paper poses problems for the turtle generator which is not based on the position of the margins ... (impacted by Shift X and Y)


i-make-robots:
Looks to me like Paper is still using Log instead of Logger.

i-make-robots:
Nice work!

PR #416, opened 2021-12-26
New look and feel of the plotter control interface

coliss86:
New lnf:
image

when clicking on the "advanced controls", the window expands it self:
image

When connecting to a robot, home and emergency stop are enabled:
image

When homing is done, the button in the panel "Draw Controls" are enabled:
image

Added an error dialog when an error is received from the printer:
image

I have added some unit test with mockito to test the SerialTransportLayer.


PPAC37:
Magnificent !

(But you give us work because it will force us to do this everywhere otherwise it will denote elsewhere :) (kidding ! good job !) .)

I said to myself that it will also need to add ToolTips and keyevent accelerators, maybe using AbstractAction ...

i-make-robots:
It is very nice and colorful.  Sadly it still has those three tabs in the middle.  Rather than delete things, what if the was a SimplePlotterControl that hid the more advanced version behind an "advanced" button?  builds quick, only has what you need on it, almost no compromise.

coliss86:
yes, this was in my mind

coliss86:
Full view:
image


Collapsed and connected:
image

coliss86:
Update:
image

i-make-robots:
Please changed "Debug controls" to "Advanced controls".  Debug makes us look unprepared :)

coliss86:
> Please changed "Debug controls" to "Advanced controls". Debug makes us look unprepared :)

ok :-)

i-make-robots:
So the new panel only works when the robot is connected.  Can I still simulate moving the Plotter when not connected through the Advanced tab?

coliss86:
> So the new panel only works when the robot is connected. Can I still simulate moving the Plotter when not connected through the Advanced tab?

I just checked on master: when the robot is not connected, the cartesian button is disabled. The same behavior is implemented. Pen up, down disengage are always enabled as master.
What am I missing ?

i-make-robots:
They are disabled until the robot is homed.

coliss86:
On master: the robot is not connected. Pen up / down and disengage are enabled.
image

i-make-robots:
Correct.  Plotter does not need to home in order to perform those actions.

i-make-robots:
"find home" should still be possible, even when simulating.  after homing, belts, weights, and pen appear on screen, cartesian driving unlock.  See current master:
master

coliss86:
I removed the second "Find home" because it was redundant to the first one.
Do you want it back ?

i-make-robots:
Marlin interface is a bare connection to the robot.

Jog interface extends it to make Cartesian driving easy, it should do all the things that a user wants.

Program interface sits between the two and provides the command buffer. It’s almost an undo system.

Plotter controls brings it all together with the most usable human interface. The friendliest most intuitive look and feel.  It hides the other three.

src/main/java/com/marginallyclever/communications/TransportLayer.java
```diff
@@ -11,17 +13,14 @@
 	 * 
 	 * @return english name of this transport layer
 	 */
-  public String getName();
+    String getName();
```
i-make-robots, line 16: no explicit access rights?

src/main/java/com/marginallyclever/communications/tcp/TCPTransportLayer.java
```diff
@@ -49,7 +50,7 @@ public NetworkSession openConnection(String connectionName) {
 	}
 
 	@Override
-	public TransportLayerPanel getTransportLayerPanel() {
-		return new TCPTransportLayerPanel(this);
+	public List<String> listConnections() {
+		return List.of("192.168.1.183:9999");
```
i-make-robots, line 54: what is this hard coded ip?

src/main/java/com/marginallyclever/convenience/ButtonIcon.java
```diff
@@ -0,0 +1,37 @@
...
+
+import javax.swing.*;
+import java.awt.*;
+
+public class ButtonIcon extends JButton {
+    private static final Logger logger = LoggerFactory.getLogger(ButtonIcon.class);
+
+    public ButtonIcon(String translationKey, String iconPath) {
+        super();
+        if (translationKey != null && !translationKey.isEmpty() && !translationKey.isBlank()) {
```
i-make-robots, line 15: Does Translator.get() fail if given a null?  That might be the place to return "no key provided" or something equivalent.

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -1,6 +1,6 @@
 package com.marginallyclever.makelangelo;
-/**
```
i-make-robots, line 2: why?

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -284,11 +285,10 @@ private void estimateTime() {
 	}
 
 	private void openPlotterControls() {
-		PlotterControls plotterControls = new PlotterControls(myPlotter,myTurtle);
-		JDialog dialog = new JDialog(mainFrame,PlotterControls.class.getSimpleName());
+		JDialog dialog = new JDialog(mainFrame, Translator.get("PlotterControls.Title"));
+		PlotterControls plotterControls = new PlotterControls(myPlotter,myTurtle, dialog);
```
i-make-robots, line 289: why does PlotterControls need dialog?  for a parent component?

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -309,30 +309,22 @@ private JMenu createToolsMenu() {
...
-			
-			if(pc != null) {
-				JButton bCapture = new JButton(Translator.get("MenuCaptureImage"));
-				bCapture.addActionListener((e)->{
-					pc.run(mainFrame,myPaper);
-				});
-		        menu.add(bCapture);
-				menu.addSeparator();
-		    } 
+
```
i-make-robots, line 312: PiCapture action can be null here and this change will fail.

src/main/java/com/marginallyclever/makelangelo/Translator.java
```diff
@@ -92,18 +92,15 @@ static private boolean doesLanguagePreferenceExist() throws BackingStoreExceptio
 	 * save the user's current language choice
 	 */
 	public static void saveConfig() {
-		logger.debug("Translator::saveConfig()");
+		logger.debug("saveConfig()");
```
i-make-robots, line 95: we have many saveConfig().  the new message is more ambiguous.

src/main/java/com/marginallyclever/makelangelo/firmwareUploader/FirmwareUploaderPanel.java
```diff
@@ -74,7 +74,7 @@ private void updateCOMPortList() {
 	}
 	
 	private String[] getListOfCOMPorts() {
-		return SerialTransportLayer.listConnections();
+		return new SerialTransportLayer().listConnections().toArray(new String[0]);
```
i-make-robots, line 77: Why is it better to have SerialTransportLayer *not* be a singleton?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConnectionButton.java
```diff
@@ -0,0 +1,128 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.swing.*;
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ConnectionButton extends JPanel {
```
i-make-robots, line 16: ConnectionButton is not as descriptive as the old name.  This class is much more than a single button.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConversationHistory.java
```diff
@@ -21,8 +25,8 @@
 	private ConcurrentLinkedQueue<ConversationEvent> inBoundQueue = new ConcurrentLinkedQueue<ConversationEvent>();
 	private JFileChooser chooser = new JFileChooser();
 
-	private JButton bClear = new JButton("Clear");
-	private JButton bSave = new JButton("Save");
+	private ButtonIcon bClear = new ButtonIcon("ConversationHistory.Clear", "/images/application.png");
+	private ButtonIcon bSave = new ButtonIcon("ConversationHistory.Save", "/images/disk.png");
```
i-make-robots, line 29: I love it.  Nice!

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
```diff
@@ -110,23 +106,30 @@ private void setupNetworkListener() {
...
 			}
 		}
 	}
 
 	private void onHearResend(String message) {
 		String numberPart = message.substring(message.indexOf(STR_RESEND) + STR_RESEND.length());
 		try {
-			int n = Integer.valueOf(numberPart);
-			if(n>lineNumberAdded-MarlinInterface.HISTORY_BUFFER_LIMIT) {
+			int n = Integer.parseInt(numberPart);
```
i-make-robots, line 131: good catch.  a different PR should check all instances of "valueOf" and replace them with parse..()

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
```diff
@@ -57,25 +60,19 @@
 	private Timer timeoutChecker = new Timer(10000,(e)->onTimeoutCheck());
 	private long lastReceivedTime;
 	
-	public MarlinInterface() {
+	public MarlinInterface(ConnectionButton connectionButton) {
```
i-make-robots, line 63: I'm not a fan of blowing up default constructors.  I'm not sure why the new PlotterControls are changing MarlinInterface, which wasn't broken and is only visible to advanced users...?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
```diff
@@ -110,23 +106,30 @@ private void setupNetworkListener() {
...
+
+			if (logger.isTraceEnabled() && !message.startsWith(STR_OK)) {
+				logger.trace("received '{}'", message.trim());
+			}
+
+			if (message.startsWith(STR_OK)) {
 				onHearOK();
-			} else if(message.contains(STR_RESEND)) {
+			} else if (message.contains(STR_RESEND)) {
+				logger.trace("received '{}'", message.trim());
```
i-make-robots, line 120: this will duplicate the previous logger.trace

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
```diff
@@ -110,23 +106,30 @@ private void setupNetworkListener() {
...
-		if(evt.flag == NetworkSessionEvent.DATA_RECEIVED) {
-			lastReceivedTime=System.currentTimeMillis();
+		if (evt.flag == NetworkSessionEvent.DATA_RECEIVED) {
+			lastReceivedTime = System.currentTimeMillis();
 			String message = ((String)evt.data).trim();
-			//logger.debug("MarlinInterface received '"+message.trim()+"'.");
-			if(message.startsWith(STR_OK)) {
+
+			if (logger.isTraceEnabled() && !message.startsWith(STR_OK)) {
+				logger.trace("received '{}'", message.trim());
```
i-make-robots, line 114: this will lose OK messages, which might indicate when connection was lost.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
```diff
@@ -208,46 +220,16 @@ public boolean getIsBusy() {
...
 	}
 
-	private void updateButtonAccess() {
-		boolean isConnected = chatInterface.getIsConnected();
-
-		bESTOP.setEnabled(isConnected);
-	}
-
-	public void setNetworkSession(NetworkSession session) {
-		chatInterface.setNetworkSession(session);
```
i-make-robots, line 241: chatInterface still exists.  so how does it get the right NetworkSession?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinPlotterInterface.java
```diff
@@ -135,7 +135,7 @@ protected void onHearAcceleration(String message) {
 			logger.debug("MarlinPlotterInterface found acceleration {}", v);
 			myPlotter.getSettings().setAcceleration(v);
 		} catch (Exception e) {
-			logger.error("M201 error: {}", message, e);
+			logger.warn("M201 error: {}", message, e);
```
i-make-robots, line 138: why the downgrade to warning?  it literally says "error" in the output.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
```diff
@@ -1,90 +1,151 @@
...
-		pane.addTab(JogInterface.class.getSimpleName(), jogInterface);
-		pane.addTab(MarlinInterface.class.getSimpleName(), marlinInterface);
-		pane.addTab(ProgramInterface.class.getSimpleName(), programInterface);
+		JTabbedPane tabbedPane = new JTabbedPane();
+		jogInterface.setPreferredSize(new Dimension(580, 300));
+		tabbedPane.addTab(Translator.get("PlotterControls.JogTab"), jogInterface);
+		tabbedPane.addTab(Translator.get("PlotterControls.MarlinTab"), marlinInterface);
+		tabbedPane.addTab(Translator.get("PlotterControls.ProgramTab"), programInterface);
+
+		CollapsiblePanel collapsiblePanel = new CollapsiblePanel(parentWindow, Translator.get("PlotterControls.AdvancedControls"), 570);
```
i-make-robots, line 55: parentWindow could be PlotterControls?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
```diff
@@ -1,90 +1,151 @@
...
+		JPanel panel = new JPanel();
+		panel.setLayout(new BorderLayout());
+		panel.add(getConnectPanel(), BorderLayout.NORTH);
+		panel.add(getDrawPanel(), BorderLayout.CENTER);
+		return panel;
+	}
+
+	private JPanel getConnectPanel() {
+		JPanel panel = new JPanel();
+		Border border = BorderFactory.createTitledBorder(Translator.get("PlotterControls.ConnectControls"));
```
i-make-robots, line 83: I don't like taking two classes and smashing them into one.  PlotterControls should delegate these responsibilities to the ChooseConnection.  

Having said that, I like how you improved the look and feel.  It's much better without a popup.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/TextInterfaceToListeners.java
```diff
@@ -15,7 +15,7 @@
 	 * 
 	 */
 	private static final long serialVersionUID = 7996257740483513358L;
-	private JTextField commandLine = new JTextField(60);
+	private JTextField commandLine = new JTextField(30);
```
i-make-robots, line 18: 60 is the maximum character limit to send to a mega2560 in a single 'send'.  This was a carefully chosen number :)

src/main/resources/languages/english.xml
```diff
@@ -577,6 +573,32 @@
 		<value>Save Gcode to file/SD</value>
 		<hint>menu or dialog title</hint>
 	</string>
+
+	<string>
+		<key>PlotterControls.Title</key>
```
i-make-robots, line 578: Thank you for cleaning these up.  Little by little we will get them all.

src/main/java/com/marginallyclever/communications/TransportLayer.java
coliss86: This method belongs to an interface which is `public`. You can't reduce the visibility of a method, and as `public` is the highest visibility, you can omit this declaration.

src/main/java/com/marginallyclever/communications/tcp/TCPTransportLayer.java
coliss86: It's taken from here: https://github.com/MarginallyClever/Makelangelo-software/blob/c391867f69137d7d5e606ce262314cb7a8683fc8/src/main/java/com/marginallyclever/communications/tcp/TCPTransportLayerPanel.java#L20

src/main/java/com/marginallyclever/convenience/ButtonIcon.java
coliss86: it throws a NullPointerException if I remember correctly. I suggest improving Translator in an other PR

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
coliss86: good question, I'll revert this modification

src/main/java/com/marginallyclever/communications/NetworkSession.java
```diff
@@ -30,6 +31,8 @@ public String getName() {
 	
 	// OBSERVER PATTERN
 
+	private List<NetworkSessionListener> listeners = new ArrayList<>();
```
i-make-robots, line 34: Why is it better?

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
coliss86: yes, it is need as a parent

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
coliss86: It is a constructor, if it fails, it will throw an exception, if it succeed the object `PiCaptureAction` will be instantiated.

src/main/java/com/marginallyclever/makelangelo/Translator.java
coliss86: When a log is printed, the package+class in printed like this: 
```
2021-12-30 22:06:51,618 DEBUG com.marginallyclever.makelangelo.Translator - saveConfig
```
or in dev:
```
2021-12-30 22:06:51,618 DEBUG c.m.makelangelo.Translator - saveConfig
```

so you know where it came from.

src/main/java/com/marginallyclever/makelangelo/firmwareUploader/FirmwareUploaderPanel.java
coliss86: It was not a real singleton, it was a static method but you can say it is the same thing :)
In the whole application, there is few class containing only static method which is a good thing. It will be useful when new unit test will be added, it will be easier to mock it.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConnectionButton.java
coliss86: It refers to the button which connect to the robot: 
image

What name do you prefer ?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
coliss86: `MarlinInterface` contained the button to connect to the robot. It is needed to link the serial transport to the interface. Now with this modification as the button is outside this interface, I have to pass it as an argument.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
coliss86: ok, I will log every time

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
coliss86: good catch

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
coliss86: OK !

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
coliss86: An event is emitted from the `ConnectionButton` and `TextInterfaceToNetworkSession` is listening for it here:
https://github.com/MarginallyClever/Makelangelo-software/blob/138fc0f42d29dd9d2f76a1588aece8e8b21c3454/src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/TextInterfaceToNetworkSession.java#L48

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinPlotterInterface.java
coliss86: yep, but this is not fatal for the application so warning is good enough. If it was an error, there will be a message to the user or something else in the catch. Here nothing is done except a log.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
coliss86: no, the `CollapsiblePanel` need a dialog box to be able to resize it when opening of collapsing it

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
coliss86: You mean having a new panel `ChooseConnectionPanel` which take care of all of this ?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/TextInterfaceToListeners.java
coliss86: Ok ! I'll revert it, didn't know :)

src/main/java/com/marginallyclever/communications/NetworkSession.java
coliss86: I moved it to make it homogen as other classes. I like this organization with the comment: `// OBSERVER PATTERN`: it makes it clear.

coliss86:
> Marlin interface is a bare connection to the robot. Jog interface extends it to make Cartesian driving easy, it should do all the things that a user wants. Program interface sits between the two and provides the command buffer. It’s almost an undo system. Plotter controls brings it all together with the most usable human interface. The friendliest most intuitive look and feel. It hides the other three.

I have added back the home button in the JogInterface: 
image

src/main/java/com/marginallyclever/communications/TransportLayer.java
i-make-robots: i still prefer explicit so there's no doubt.

src/main/java/com/marginallyclever/convenience/ButtonIcon.java
i-make-robots: What was the other PR #?

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -455,9 +446,7 @@ public void openLoadFile(String filename) {
 		logger.debug("Loading file {}...", filename);
 		try {
 			LoadFilePanel loader = new LoadFilePanel(myPaper,filename);
-			loader.addActionListener((e)->{
-				setTurtle((Turtle)(e).getSource());
-			});
+			loader.addActionListener((e)-> setTurtle((Turtle)(e).getSource()));
```
i-make-robots, line 449: in general these are multi-line because it's easier to set a break point.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConnectionButton.java
i-make-robots: Ok, if it's just the button.  I see ChooseConnection.java was deleted so maybe I'm just confused.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConnectionButton.java
```diff
@@ -0,0 +1,135 @@
...
+	private static final Logger logger = LoggerFactory.getLogger(ConnectionButton.class);
+	
+	private static final long serialVersionUID = 4773092967249064165L;
+	@Deprecated
+	public static final int CONNECTION_OPENED = 1;
+	@Deprecated
+	public static final int CONNECTION_CLOSED = 2;
+	
+	private ButtonIcon bConnect = new ButtonIcon("ButtonConnect", "/images/connect.png");
```
i-make-robots, line 32: This is the connect button.  The rest is part of the job of choosing a connection.  That's why I preferred "ChooseConnection".

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConversationHistory.java
```diff
@@ -49,11 +53,11 @@ public ConversationHistory() {
 	
 	private JToolBar getToolBar() {
 		JToolBar bar = new JToolBar();
-		bar.setRollover(true);
+		bar.setFloatable(false);
```
i-make-robots, line 56: part of me doesn't want to take away the power to float.  The rest of me doesn't want to support saving preference for float location, or the headache of tech support when a user has moved the floating toolbar.  So this is good.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/JogInterface.java
```diff
@@ -82,32 +85,35 @@ public void findHome() {
...
 		bar.add(penDown);
 		bar.addSeparator();
 		bar.add(toggleEngageMotor);
-		
-		penUp.addActionListener((e)-> myPlotter.raisePen());
-		penDown.addActionListener((e)-> myPlotter.lowerPen());
-		findHome.addActionListener((e)-> myPlotter.findHome());
-		toggleEngageMotor.addActionListener((e)-> onToggleEngageMotorAction());
+
+		home.addActionListener(e -> myPlotter.findHome());
```
i-make-robots, line 101: there used to be a "goHome" button.  keeping it called "findHome" is less ambiguous.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
i-make-robots: Ok, when is the button access updated if the updateButtonAccess method is deleted?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
i-make-robots: Yeah.  there used to be a ChooseConnection.  half of it became ConnectionButtonand half of it became ... like, what is connectionComboBox here for if there's also `ConnectionButton.connectionComboBox`?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConnectionButton.java
i-make-robots: It's not just the button.  see comments elsewhere.

src/main/java/com/marginallyclever/communications/TransportLayer.java
coliss86: OK reverted

src/main/java/com/marginallyclever/convenience/ButtonIcon.java
coliss86: it is not yet developed

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConnectionButton.java
coliss86: I renamed it to ChooseConnection

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ConnectionButton.java
coliss86: ok, it is now done

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/JogInterface.java
coliss86: Ok renamed

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinInterface.java
coliss86: In PlotterControls when the connection is opened
https://github.com/coliss86/Makelangelo-software/blob/60990f77f3f660c5fb058b25e5d6a9f6f471e666/src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java#L228

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
coliss86: You're right, renaming `ConnectionButton` to `ChooseConnection` and your comments show me that `ChooseConnection` is not just a button but a panel.
I moved the `combobox` into `ChooseConnection`. It's better now ! Thanks

PR #420, opened 2021-12-27
there is no need of the JFrame ( in testColor use directly panel (a J…

PPAC37:
…component as parent) )
https://github.com/MarginallyClever/Makelangelo-software/issues/419


i-make-robots:
If the frame fails to create each test after will also fail.  How does this solve for all cases?

PPAC37:
For me, a JComponent can be instantiated without being added to a JFrame. We can test it without JFrame.
This will reduce the scope of the test because ultimately the component will not be displayed. 
So only for functional logic tests, but it seems to me that this is the purpose of the Tests.

PR #422, opened 2021-12-28
Resurrect collapsiblePanel

coliss86:
Full expanded:
image

collapsed:
image

I'll use it for the plotter interface

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/CollapsiblePanel.java
+++ b/src/main/java/com/marginallyclever/makelangelo/CollapsiblePanel.java
@@ -15,11 +15,11 @@ public class CollapsiblePanel extends JPanel {
     private String title = "";
     private final TitledBorder border;
     private final JPanel innerPannel;
-    private final JFrame parentFrame;
+    private final Window parentWindow;
     private Dimension previousDimension;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/marginallyclever/makelangelo/CollapsiblePanel.java
```diff
@@ -127,17 +127,17 @@ protected void toggleVisibility(boolean visible) {
...
+            previousDimension = parentWindow.getSize();
             int height = previousDimension.height - innerPannel.getHeight();
             Dimension toggle = new Dimension(previousDimension.width, height);
-            parentFrame.setPreferredSize(toggle);
+            parentWindow.setPreferredSize(toggle);
         }
-        parentFrame.validate();
-        parentFrame.repaint();
-        parentFrame.pack();
+        parentWindow.validate();
```
i-make-robots, line 138: revalidating this item cause the parent to automatically revalidate?  These may be overreach.

coliss86: I used it order to redraw the parent interface when the panel is collapsed. otherwise, the parent window stays at the maximum height

PR #436, opened 2022-01-10
Introduced a Machines enum in order to centralize all machine type

coliss86:
Hi,

I refactored the `PlotterRendererFactory` in a enum `Machines` in order to externalize this list from the renderer. For now, only the render change from one machine to another but some characteristics will be added in the future like with/without endstop, speed ...

The purpose of an enum is a list a element which cannot be changed at runtime and can be validated at compiled time.

In the future, new attribute will be added to the `Machines` enum.

What do you think ?


i-make-robots:
proposition: we've already covered all the types of plotters that will exist for a very long time.
If that is true, an enum is fine.

If that is not true, it is likely that later there will be a system to download other machines from their github repositories, same way arduino installs new libraries.  In that case enum is not going to work.  I know that I already want to do something like this in Robot Overlord because robot files are BIG and don't need to be in the JAR.

scara plotters are not in the current system.  advanced graphics for better look & feel are not in the current system.  coreXY plotters are not in the current system.  IDK that they need to be....

PR #442, opened 2022-01-11
Move macOS look and feel to setSystemLookAndFeel()

coliss86:
Here a small refactor, I just discovered the `setSystemLookAndFeel()` method, so I moved the os detection for the menu bar here.

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
+++ b/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
@@ -102,7 +102,7 @@ public final class Makelangelo {
 	private Plotter myPlotter;
 	private Paper myPaper = new Paper();
 	private Turtle myTurtle = new Turtle();
-	private boolean isMacOS = false;
+	private static boolean isMacOS = false;
 
 	private TurtleRenderFacade myTurtleRenderer = new TurtleRenderFacade();
 	private PlotterRenderer myPlotterRenderer = Machines.MAKELANGELO_5.getPlotterRenderer();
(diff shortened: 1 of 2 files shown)


PPAC37:
Thanks for the corrections with the logger.
My apologies, for the mess.

coliss86:
Don't worry :)

coliss86:
Marking as draft, I found another enhancement

coliss86:
Finally it's ready. I didn't manage for now to do what I want

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -180,6 +175,11 @@ private static void setSystemLookAndFeel() {
 	        	UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
 	        } catch (Exception e) {}
 		}
+		String os = System.getProperty("os.name").toLowerCase(Locale.ENGLISH);
```
i-make-robots, line 178: this seems pretty self-contained and could probably be in a separate method.

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -102,7 +102,7 @@
 	private Plotter myPlotter;
 	private Paper myPaper = new Paper();
 	private Turtle myTurtle = new Turtle();
-	private boolean isMacOS = false;
+	private static boolean isMacOS = false;
```
i-make-robots, line 105: why is this static?

coliss86: because `setSystemLookAndFeel()` is static

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
coliss86: ok, done

PR #443, opened 2022-01-11
Save and load machine style

coliss86:
Load the machine type at startup and save it when selecting it in the Robot menu.
If the settings was never saved, the default machine is `MAKELANGELO_5`.

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/machines/Machines.java
+++ b/src/main/java/com/marginallyclever/makelangelo/machines/Machines.java
@@ -5,15 +5,18 @@ import com.marginallyclever.makelangelo.plotter.plotterRenderer.*;
 import java.util.Arrays;
 
 public enum Machines {
-    MAKELANGELO_5(new Makelangelo5()),
-    MAKELANGELO_3_3(new Makelangelo3_3()),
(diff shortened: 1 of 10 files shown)


src/main/java/com/marginallyclever/makelangelo/machines/Machines.java
```diff
@@ -5,25 +5,32 @@
 import java.util.Arrays;
 
 public enum Machines {
-    MAKELANGELO_5(new Makelangelo5()),
-    MAKELANGELO_3_3(new Makelangelo3_3()),
-    MAKELANGELO_CUSTOM(new MakelangeloCustom()),
-    CARTESIAN(new Cartesian()),
-    ZARPLOTTER(new Zarplotter());
+    MAKELANGELO_5("Makelangelo 5", new Makelangelo5()),
```
i-make-robots, line 8: this new method has the Machines enum String and the PlotterRender.getName() as the same value.  I hope they never change?

coliss86: `PlotterRender.getName()` is replaced by this enum String :)

PR #444, opened 2022-01-12
Zoom on the machine at startup

coliss86:
When selecting a small paper size and restarting the app, the viewport is zoomed on paper, and the user is confused.

Here is an example with A4.

Before:
image

After:
image


i-make-robots:
so if I change the paper to A2, quit and restart... does everything look right?

coliss86:
I only changed the zoom and not `myPaper` size, but I used the A2 size as the default zoom.
Here is an example with A2 with this feature. I select A2: 
image

Quit, restart: 
image

coliss86:
OK :)

i-make-robots:
tested, seems fine.

PR #445, opened 2022-01-12
Add translation to paper settings and add missing translation to robot menu

coliss86:
I also changed `<->` to Landscape

image


i-make-robots:
Maybe "Shift X" should be "Shift left" and "Shift Y" should be "Shift up".  It's more clear to non-programmers?

coliss86:
Yep, why not

PR #449, opened 2022-01-13
Fix #446: the robot control window is not resizable anymore

coliss86:
Fix the size of the windows to avoid the problem described in #446
> Also, show advanced controls, then drag window bottom downward. tabs to not resize to fill.

PR #450, opened 2022-01-13
Disable / enable the buttons in the advanced based on the connection status

coliss86:
Must be merged after #447 and #449.

When opening the Robot control: 
image

When the button connect is clicked: 
image

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/JogInterface.java
+++ b/src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/JogInterface.java
@@ -29,7 +29,10 @@ public class JogInterface extends JPanel {
 	private Plotter myPlotter;
 	private CartesianButtons bCartesian = new CartesianButtons();
 	private JButton toggleEngageMotor;
-	
+	private JButton findHome;
+	private JButton penUp;
+	private JButton penDown;
+
 	public JogInterface(Plotter plotter) {
 		super();
 		myPlotter=plotter;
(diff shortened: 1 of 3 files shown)


src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/JogInterface.java
```diff
@@ -72,23 +75,32 @@ public JogInterface(Plotter plotter) {
...
 		});
-		updateButtonStatus();
+		updateButtonStatusWithPlotter();
+		updateButtonsStatus(false);
 	}
 	
 	public void findHome() {
 		myPlotter.findHome();
-		updateButtonStatus();
+		updateButtonStatusWithPlotter();
```
i-make-robots, line 87: this will occur when the plotter says it has found home.  redundant here?

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
```diff
@@ -93,12 +107,7 @@ private JPanel getConnectPanel() {
 		Border border = BorderFactory.createTitledBorder(Translator.get("PlotterControls.ConnectControls"));
 		panel.setBorder(border);
 		panel.add(chooseConnection);
-		chooseConnection.addListener(e -> {
-			switch (e.flag) {
-				case NetworkSessionEvent.CONNECTION_OPENED -> updateButtonStatusOnConnect();
```
i-make-robots, line 98: no `break;`?

coliss86: this is a new case syntax introduced in java 14 : https://betterprogramming.pub/a-look-at-the-new-switch-expressions-in-java-14-ed209c802ba0

PR #455, opened 2022-01-17
Traduction debug lost values due to not unique key (To start a discution about Traduction maintenance)

PPAC37:
Hello !
Things to review in the translations.
Both in code and .xml files
Currently some translation values ​​are lost because

there are several times the same translation key in the .xml files
ex:
```<key>...</key>```
* Left
* Right
* MenuSaveFile
* spraypaintToolTest
* MenuResetMachinePreferences
* Menu Log
* PaperSize
* voronoiStipplingCellCount
* voronoiStipplingDotMin
* NotConnected
* FileTypeImage
* Open

 or because there is more than one translation ```<value>``` (the current code only use the first found) for a ```<key>```
https://github.com/MarginallyClever/Makelangelo-software/blob/03e9f11afac3e7badf82b8851d435d81fc305c14/src/main/resources/languages/german.xml#L1059-L1063

There is also the question of maintaining the .xml translation files when changing the value of a key.

A list of the keys (when renaming a key) of the translation files should be kept up to date.
This would make it easier to maintain the other translation files. (to update a new translation file created on an old version ...)

example in 
https://github.com/MarginallyClever/Makelangelo-software/commit/9463279394b671ae0b3d9475f9678d283c026fac
```
<key>PaperWidth</key> -> <key>PaperSettings.PaperSize</key>
<key>PaperHeight</key> -> <key>PaperSettings.PaperHeight</key>
<key>OpenControls</key> -> <key>RobotMenu.OpenControls</key>
<key>SaveGCode</key> -> <key>RobotMenu.SaveGCode</key>
```
But the other translation files seems not have been modified/impacted...

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/Translator.java
+++ b/src/main/java/com/marginallyclever/makelangelo/Translator.java
@@ -145,7 +145,7 @@ public final class Translator {
 			URL pathFound = Translator.class.getClassLoader().getResource(path);
 			logger.debug("path found: {}", pathFound);
 			try (InputStream s = pathFound.openStream()) {
-				languageContainer.loadFromInputStream(s);
+				languageContainer.loadFromInputStream(s, pathFound.toString());
 			} catch (IOException ie) {
 				logger.error(ie.getMessage());
 			}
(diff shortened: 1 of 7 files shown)


coliss86:
nice, don't forget to mark this PR as draft because there is many comment about debug and wip

PPAC37:
No longer of interest.
Has been partially implemented via other PR.

the code to generate an XML file will surely be used for another PR.

Being in conflict I close this request.

PR #456, opened 2022-01-17
Search in the code traduction get to find the key ( As a draft )

PPAC37:
A Draft to search in the src file the arguments used in Traduction.get(...)
Badly done. 
But to open/continue a discution about some tools / good practice to maintaine traductions .xml files.


coliss86:
Nice idea, you could go one step further and add it this as a Unit Test: it will scan all the source and fails if a translation is missing while displaying the culprit. Thus, it would be executed every time by the github action and the developer could fix the code asap ==> no need a JFrame to display the result.
WDYT ?

PPAC37:
> you could go one step further and add it this as a Unit Test: it will scan all the source and fails if a translation is missing while displaying the culprit. Thus, it would be executed every time by the github action and the developer could fix the code asap ==> no need a JFrame to display the result.
> WDYT ?

Yes that seems like a good idea.
Provided that I refine again to retain only simple text strings.
This will remain a partial check, if the argument(s) to Traduction.get() are anything other than a simple text string, it will then be difficult to get the value of the key...

Also, I tell myself that it can also go wrong (my code probably has a lot of possible flaws) ... and that it may be too violent to invalidate the build for missing translations.


I was more on the idea of ​​allowing to modify / rename the key values via the JTable / TableModel to impact each translation file (eventually see in the code) (and take the opportunity to memorize the key renaming carried out by this method to have a knowledge base to possibly later update more easily new translation file which will be based on an old file (whose key has possibly been renamed)...)

coliss86:
From my experience, a tool that you manually run to check codes is ran few times and forgot. That is why I was suggesting a unit test. This test can also be refine later if needed.

PPAC37:
* [ ] logger
* [ ] Unit Test
* [ ] cleanning
* [ ] javadoc

src/main/java/com/marginallyclever/util/FindAllTraductionGet.java
```diff
@@ -0,0 +1,278 @@
...
+		    .filter(Files::isRegularFile)
+		    .filter(Files::isReadable)
+		    .map(Path::toFile)
+		    .filter(f -> f.getName().endsWith(fileNameEndsWithSuffix))
+		    .map(File::toPath)
+		    .collect(Collectors.toList());
+	}
+	if (debugListFiles) {
+	    if (result == null) {
+		logger.debug(String.format("listFiles(\"%s\",\"%s\")=null", path, fileNameEndsWithSuffix));
```
coliss86, line 270: the `logger` API supports format Strings. You can convert this line with:
```
logger.debug("listFiles(\"{}\",\"{}\")=null", path, fileNameEndsWithSuffix);
```

PPAC37: Thank you !
And sorry, I should have done it from the start.

src/main/java/com/marginallyclever/util/FindAllTraductionGet.java
```diff
@@ -0,0 +1,278 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+ */
+package com.marginallyclever.util;
+
+import com.marginallyclever.convenience.CommandLineOptions;
+import com.marginallyclever.convenience.log.Log;
```
coliss86, line 21: This import is not needed, the one below is used instead: 
```
import org.slf4j.Logger;
```

src/main/java/com/marginallyclever/util/FindAllTraductionGet.java
```diff
@@ -0,0 +1,278 @@
...
+
+    /**
+     * Try to search a src java project for a specific pattern. like
+     * <code>Traduction.get(...)</code> in all .java file.
+     *
+     * @param args
+     */
+    public static void main(String[] args) {
+
+	Log.start();
```
coliss86, line 65: Don't forget to indent properly the code :)

src/test/java/com/marginallyclever/makelangelo/TranslatorTests.java
```diff
@@ -49,4 +59,46 @@ public void changeLanguageTest() {
...
+		//Pre requis
+		Translator.start();
+		
+		String baseDirToSearch = "src" + File.separator + "main" + File.separator + "java";
+		System.out.printf("PDW=%s\n", new File(".").getAbsolutePath());
+		File srcDir = new File(".", baseDirToSearch);
+		try {
+		    System.out.printf("srcDir=%s\n", srcDir.getCanonicalPath());
+		} catch (IOException ex) {
+		    ex.printStackTrace();
```
coliss86, line 75: If you don't know what to do with the exception, let it flow. You are in a test, thus, the runner will fails with a beautiful error message

src/test/java/com/marginallyclever/makelangelo/TranslatorTests.java
```diff
@@ -49,4 +59,46 @@ public void changeLanguageTest() {
...
+//			System.out.printf(" used in : \"%s\" line %d\n", tr.pSrc, tr.lineInFile);
+//		    }
+//		}
+		FindAllTraductionXMLGenerator.generatePartialXmlFileWithMissingKey(groupIdenticalMissingKey);
+		
+		// validate or not the test. (succes if no missing keys found)
+		assertTrue(groupIdenticalMissingKey.isEmpty(), "Some traduction missing keys !?.");
+		
+	    } catch (Exception e) {
+		e.printStackTrace();
```
coliss86, line 101: If the code fails for whatever reasons, the test will be mark as Green, aka : OK. Is it what you want ?

src/test/java/com/marginallyclever/makelangelo/TranslatorTests.java
PPAC37: Thanks !

src/test/java/com/marginallyclever/makelangelo/TranslatorTests.java
PPAC37: Yes you right !  my bad ... have to do some change ...

PPAC37:
As @coliss86  has made it much simpler and cleaner to detect missing keys in https://github.com/MarginallyClever/Makelangelo-software/pull/502
 and as the other functionalities (JTable / Table Model of translations, partial language .xml file generator / template ) seem non-essential or to be separated.
I therefore close this draft request.

PR #466, opened 2022-01-20
Add unit test for PlotterSettings

coliss86:
Fix #440

PR #467, opened 2022-01-20
Fix unit test for PlotterSettings introduced in #466

coliss86:


coliss86:
I don't understand why I can't see the workflow ran from my branch, do you have an "Approve and run" button ? like described here : https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

coliss86:
I hope #468 will enable the ability to view the result of the workflow in the PR (and then, you will not merge a PR which breaks unit test ;) )

PR #471, opened 2022-01-23
💄 Refresh button disabled when the connect button is clicked

coliss86:
the button was still enabled when the robot was connected, now it is fixed:
image


i-make-robots:
Nice.

PR #477, opened 2022-01-26
InfillTurtleAction fix to update the preview sliver last max limite

PPAC37:
to view the possible infil created ...

As in InfillTurtelAction, myMakelangelo.setTurtle indirectly call myTurtleRenderer.setTurtle(turtle); :
https://github.com/MarginallyClever/Makelangelo-software/blob/94745b3e8154a335e397d5a0c5091d6e1e6d78b6/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java#L974-L976

and so the last slider position is updated. ( setLast(size);  line 95 )
https://github.com/MarginallyClever/Makelangelo-software/blob/94745b3e8154a335e397d5a0c5091d6e1e6d78b6/src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFacade.java#L89-L96

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/makeArt/InfillTurtleAction.java
+++ b/src/main/java/com/marginallyclever/makelangelo/makeArt/InfillTurtleAction.java
@@ -27,6 +27,8 @@ public class InfillTurtleAction extends AbstractAction {
 		Turtle t = myMakelangelo.getTurtle();
 		try {
 			t.add(infill.run(t));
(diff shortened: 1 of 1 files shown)


i-make-robots:
It would be better if classes that change a turtle send an event for listeners.  Then they would not have to know about or include Makelangelo.

i-make-robots:
Good catch!

PR #496, opened 2022-01-29
✅ Enhance PlotterSettingsTest

coliss86:
This what I was trying to explain in https://github.com/MarginallyClever/Makelangelo-software/pull/481#discussion_r794288189

### DIFF
--- a/src/test/java/com/marginallyclever/makelangelo/plotter/settings/PlotterSettingsTest.java
+++ b/src/test/java/com/marginallyclever/makelangelo/plotter/settings/PlotterSettingsTest.java
@@ -41,22 +41,13 @@ public class PlotterSettingsTest {
         plotterSettings.setDrawFeedRate(21);
         plotterSettings.setAcceleration(22);
         plotterSettings.setHardwareName("TestRobot");
-        
-        int a = plotterSettings.getBlockBufferSize()+1;
-        int b = plotterSettings.getSegmentsPerSecond()+1;
-        double c = plotterSettings.getMinSegmentLength()+1; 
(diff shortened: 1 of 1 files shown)


i-make-robots:
if the default PlotterSettings are changed they *could* match one of these hard-coded test values.  Then the test would do nothing, which is the same thing as fail silently.  It's an extremely rare edge case ...which means it will be ages before it happens and ages more before we catch it.  My method guarantees it cannot happen.

coliss86:
I don't totally agree: when instantiating a new object with getters and setters, it's very common to have 0 as default values for numbers like `PaperSize`. With this scenario and with the code which was written like this:
```
        int a = plotterSettings.getBlockBufferSize()+1;
        int b = plotterSettings.getSegmentsPerSecond()+1;
```
`a` and `b` will both be equals to `1`.

The purpose of this test is test whether the storing and loading values are ok.
If the keys used to store value in the PreferenceHelper are wrong, you could end up with `blockBufferSize` stored in `segmentsPerSecond` and vice versa and the tests won't be able to detect it. This is why I use different values for each attributes.

What you said is also true: "if the default PlotterSettings are changed they could match one of these hard-coded test values".

A better solution wold be a mix of both solution: instead of hard coding the expected value, generate a random value, store it in a local variable and store in the tested object. The assertion will compare the local variable and the read value.

i-make-robots:
> A better solution wold be a mix of both solution: instead of hard coding the expected value, generate a random value, store it in a local variable and store in the tested object. The assertion will compare the local variable and the read value.

Sounsd good.

PR #497, opened 2022-01-29
💄 Open a file chooser directly when clicking on "Open file..."

### LINKED ISSUE #493: Files opening
**Is your feature request related to a problem? Please describe.**
When opening a file, you you have to click twice to have the choose file dialog : In the menu and on the popup that just appeared.
In gimp or inkscape, when you click on the "Open..." you can directly select the file.

Moreover, in gimp, by default the file filter is set to every type of file supported. here, you have to select the file type in svg for example and then you can select the file.

Besides, in the case of svg files, the LoadFilePanel is empty, so I think it is not needed.

**Describe the solution you'd like**
I would like that the LoadFilePanel don't show before the selection file window.
I also would like that the selection filter match all type of file, eg: dxf, svg, png jpeg... and be selected by default

**Describe alternatives you've considered**
none

coliss86:
Fix #493


src/main/resources/languages/english.xml
```diff
@@ -442,10 +442,15 @@
 	</string>
 
 	<string>
-		<key>FileTypeImage</key>
+		<key>FileChooser.FileTypeImage</key>
 		<value>Image (jpg/bmp/png/gif)</value>
 		<hint>file open dialog</hint>
 	</string>
+	<string>
```
PPAC37, line 449: If possible when renaming a `<key>` (if the `<value>` is not modified then as the translations remain valid, and if already done) also rename the `<key>` in other languages. xml who have it.

But if the `<value>` changes then the translation will surely no longer be suitable and therefore there possibly add a `<!-- translation value to review-->` in the other .xml files of the other languages ...

src/main/java/com/marginallyclever/makelangelo/makeArt/io/LoadFilePanel.java
```diff
@@ -46,6 +44,9 @@ public LoadFilePanel(Paper paper,String filename) {
 		setLayout(new BorderLayout());
 		add(getFileSelectionPanel(filename),BorderLayout.NORTH);
 		add(mySubPanel,BorderLayout.CENTER);
+
+		fc = new FileChooser(this);
```
PPAC37, line 48: It would be great if there was only one instance of FileChooser (because there is already one in the Makelangelo calss) ... to unify reopen on the last path ... but I don't see an easy way ...

PPAC37:
Tested and except for the disjunction of the last load path due to two instances of the FileChooser, I find it super cool !

src/main/java/com/marginallyclever/makelangelo/makeArt/io/LoadFilePanel.java
coliss86: Good suggestion, I have implemented it by using `PreferencesHelper`

src/main/resources/languages/english.xml
coliss86: Good point, done !

i-make-robots: if the value changes it could simply be a more accurate translation.  Probably there is a way to prepare a release candidate and then diff the language file with the last official release to see which translations have changed.

i-make-robots:
So far this seems to work.  I am glad to see the process of loading art become more intuitive.
in ancient times there was a reason I had to put them all through the load dialog and now I don't remember what it was.  Maybe I was thinking of adding flip/resize stuff there, too, and then decided against it for being repetitive.

If you want to load vectors and *never show the LoadFilePanel*.... I'm ok with that.

coliss86:
Still at least one issue: when loading a second file from the preview panel, the panel is not cleared
image

Issue #488, opened 2022-01-28
"CMYK spirals" and "CMYK" artwork: Pen stays down when asking to change to next color cyan, magenta, black


Keepthefunk:
**Steps to make the bug happen**
1. Open any Image file and create artwork with "Spiral CMYK" or "CMYK"
2. click Tools - simplify
3. Plot Artwork, starting with yellow pen

**What was supposed to happen**
when yellow component is done, the plotter should _lift the pen_ and ask for the next color, cyan

**What happens**
When yellow component is done, plotter asks for next color but the pen stays down, bleeding ink in to the paper.
Same happens when asking for the next colors magenta and black. When the last color black is done, the pen lifts up fine. Should also lift when asking for cyan, magenta and black.

**Platform (please complete the following information):**
 - Version: [e.g. 7.31.0.0]
 - OS: [Windows]
 - Robot model (if any): Custom with RAMPS and Marlin-polargraph-bugfix-2.0.x

Picture: Cyan pen was down when plotter stopped and asked for magenta. Cyan bleeding in to the paper.
![pen down while waiting](https://user-images.githubusercontent.com/83095100/151619746-d723dc6d-c765-475b-854e-35eaa95cc1cb.jpg)

Keepthefunk:
The .log file is full of "Error: Line number is not last line number+1" errors. If you need it, I can add the file

i-make-robots:
I think I understand the problem already.  Thank you. :)

i-make-robots:
gcode before simplify:
```
; 2022-01-28 at 16:01:28 PST
G28
M0 Ready yellow and click
G0 X-176.111 Y-265.577
M280 P0 S25 T50
G1 X-175.305 Y-266.169 F3000.0
G1 X-174.498 Y-266.759 F3000.0
M280 P0 S90 T50
G0 X-174.363 Y-261.896
M280 P0 S25 T50
```
gcode after simplify:
```
; 2022-01-28 at 16:00:44 PST
G28
M0 Ready black and click
G0 X-176.111 Y-265.577
M280 P0 S25 T50
G1 X-175.305 Y-266.169 F3000.0
M280 P0 S90 T50
G0 X-176.111 Y-265.577
M280 P0 S25 T50
G1 X-175.305 Y-266.169 F3000.0
M0 Ready yellow and click
M280 P0 S90 T50
```
note the pen up has moved to immediately AFTER the M0.

PPAC37:
[...]

PR #498, opened 2022-01-29
Fix converter cmyk.java passes value should not be lower than 1

PPAC37:
Partial solutions for another ConverterCMYK issue.
as explaine in https://github.com/MarginallyClever/Makelangelo-software/issues/488#issuecomment-1025005938

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/makeArt/imageConverter/Converter_CMYK.java
+++ b/src/main/java/com/marginallyclever/makelangelo/makeArt/imageConverter/Converter_CMYK.java
@@ -16,7 +16,7 @@ import org.slf4j.LoggerFactory;
  */
 public class Converter_CMYK extends ImageConverter {
 	private static final Logger logger = LoggerFactory.getLogger(Converter_CMYK.class);
-	static protected int passes=1;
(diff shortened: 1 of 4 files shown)


Keepthefunk:
first test positive: pen lifts up after yellow. Still doing more tests, will keep you informed

src/main/java/com/marginallyclever/makelangelo/makeArt/imageConverter/Converter_CMYK.java
```diff
@@ -34,9 +34,15 @@ public int getPasses() {
...
+	 * Passing a value lower than 1 set passes value to 1.
+	 * @param value 
+	 */
 	public void setPasses(int value) {
-		if(passes<1) passes=1;
-		passes=value;
+		if(value<1) 
+		    passes=1;
+		else 
+		    passes=value;
```
i-make-robots, line 45: passes=(int)Math.max(1,value)?  your way is good, too.  thank you for adding javadoc.

Keepthefunk:
Pen lifts up before stopping and asking for next pen now. THANKS! This works now so I closed #488

PR #499, opened 2022-01-30
💄 use java.awt.Desktop to enhance the user experience on mac os

coliss86:
Until now, the native menu Makelangelo on macos was not bonded to the app. When clicking on the "about", it will not show the about of Makelangelo but one built in the jvm. 
image

When quitting the app from the same menu, the confirmation was not displayed. Only when closing the window with the red button is the left corner triggers it.

It uses the [java.awt.Desktop API](https://docs.oracle.com/javase/9/docs/api/java/awt/Desktop.html) introduced in java 6 to make it works.

Here is what looks like the dialog box on the built in menu: 
image


i-make-robots:
nice.

Issue #479, opened 2022-01-26
Open and close connection: popup error


coliss86:
**Steps to make the bug happen**
1. Start Makelangelo
2. Go to the plotter controls, connect to the robot and home it => everything works fine
3. Close the plotter controls
4. Load a SVG
5. Open the plotter controls
6. Home it, the following dialog is displayed:
![image](https://user-images.githubusercontent.com/23615562/151173548-97576aa1-d921-42c6-90b1-33cbc2c71c5a.png)
After then, the robot do not respond to anything

**What was supposed to happen**
No Error

**Platform (please complete the following information):**
 - Version: master
 - OS: OS x

**Log file**
I have removed the svg parsing from the log
```
2022-01-26 14:39:16,288 DEBUG c.m.makelangelo.Translator - starting translator... 
2022-01-26 14:39:16,296 DEBUG c.m.makelangelo.Translator - Default language = English 
2022-01-26 14:39:16,367 DEBUG c.m.makelangelo.Translator - loadConfig: User Preference Node: /DrawBot/Language=English 
2022-01-26 14:39:16,668 DEBUG c.m.makelangelo.paper.Paper - loadConfig() before Paper Width = 420.00 Height = 594.00 (left=-210.00 right=210.00 top=297.00 bottom=-297.00) centerShift(X= 0.00 Y= 0.00) color (255,255,255) 
2022-01-26 14:39:16,669 DEBUG c.m.makelangelo.paper.Paper - loadConfig() after Paper Width = 210.00 Height = 297.00 (left=-105.00 right=105.00 top=148.50 bottom=-148.50) centerShift(X= 0.00 Y= 0.00) color (255,255,255) 
2022-01-26 14:39:17,672 DEBUG c.m.makelangelo.Makelangelo - Locale=en_GB 
2022-01-26 14:39:17,672 DEBUG c.m.makelangelo.Makelangelo - Headless=N 
2022-01-26 14:39:17,672 DEBUG c.m.util.PropertiesFileHelper - makelangelo.version=7.31.0 
2022-01-26 14:39:17,675 DEBUG c.m.makelangelo.Makelangelo - machine style: MAKELANGELO_CUSTOM 
2022-01-26 14:39:17,688 DEBUG c.m.makelangelo.Makelangelo - Starting robot... 
2022-01-26 14:39:17,690 DEBUG c.m.makelangelo.Makelangelo - Starting virtual camera... 
[...]

PR #500, opened 2022-01-30
:bug: #479 cleanup the listener instead of expanding scope for PlotterControl.

i-make-robots:


coliss86:
Unfortunately, it will not work because the registration of the listener is not done in the `onConnect()` but in the constructor. Thus, if you disconnect the robot and reconnect it without closing the window, `MarlinPlotterInterface` will never get news from the `Plotter`.

i-make-robots:
...have you tested it?

coliss86:
Perfect ! works like a charm, simpler than https://github.com/MarginallyClever/Makelangelo-software/pull/494

PR #502, opened 2022-01-30
✅ Add unit test to find missing translation in english

coliss86:
ping @PPAC37, this PR is heavily inspired from your https://github.com/MarginallyClever/Makelangelo-software/pull/456

Here what happened when a translation is missing: 
```
21:48:01.532 [main] INFO  com.marginallyclever.makelangelo.MissingTranslationTest - translations missing:
21:48:01.534 [main] INFO  com.marginallyclever.makelangelo.MissingTranslationTest -   src\main\java\com\marginallyclever\makelangelo\makeArt\imageConverter\Converter_CannyEdge.java:20: CannyEdgeConverterName
21:48:01.535 [main] INFO  com.marginallyclever.makelangelo.MissingTranslationTest -   src\main\java\com\marginallyclever\makelangelo\makeArt\imageConverter\Converter_ColorFloodFill.java:45: RGBFloodFillName
...
```

But first, I have to do another PR the fix missing translation :)

### DIFF
--- a/src/test/java/com/marginallyclever/makelangelo/MissingTranslationTest.java
+++ b/src/test/java/com/marginallyclever/makelangelo/MissingTranslationTest.java
@@ -1,0 +1,130 @@
+package com.marginallyclever.makelangelo;
+
+import com.marginallyclever.util.PreferencesHelper;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MissingTranslationTest {
+
+    private static final Logger logger = LoggerFactory.getLogger(MissingTranslationTest.class);
+
+    private Pattern pattern = Pattern.compile("Translator\\s*\\.\\s*get\\s*\\(\"(?<key>[^\\)]*)\"\\)");
+
+    @BeforeAll
+    public static void init() {
+        PreferencesHelper.start();
+        Translator.start();
+    }
+
+    @Test
+    public void findMissingTranslations() throws IOException {
+        List<String> results = new ArrayList<>();
+        File srcDir = new File("src" + File.separator + "main" + File.separator + "java");
+
+        List<File> files = listFiles(srcDir.toPath(), ".java");
+        // search in the files ...
+        files.forEach(file -> {
+            try {
+                searchInAFile(file, results);
+            } catch (IOException e) {
+                logger.warn("Can read file {}", file, e);
+            }
+        });
+
+        if (results.size() != 0) {
+            logger.info("translations missing:");
+            for (String result: results) {
+                logger.info("  {}", result);
+            }
+        }
+        assertEquals(0, results.size(), "Some translations are missing");
+    }
+
+    @Test
+    public void verifyThatMatcherWorks_results() throws IOException {
+        List<String> results = new ArrayList<>();
+
+        // matches.txt contains few entries with translation and one without. The list should not be empty
+        searchInAFile(new File("src/test/resources/translator/matches.txt"), results);
+
+        assertEquals(List.of("src/test/resources/translator/matches.txt:4: unknownKey"), results);
+    }
+
+    @Test
+    public void verifyThatMatcherWorks_noResult() throws IOException {
+        List<String> results = new ArrayList<>();
+
+        searchInAFile(new File("src/test/resources/translator/no-matche.txt"), results);
+
+        // no-matche does not contains any code matching the translator. The list must be empty
+        assertEquals(0, results.size());
+    }
+
+    public void searchInAFile(File file, List<String> results) throws IOException {
+        int lineNb = 1;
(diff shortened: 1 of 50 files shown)


PPAC37:
Really better as it is simpler than my https://github.com/MarginallyClever/Makelangelo-software/pull/456

as you are basing yourself on my code, this also returns the Traduction.get in comment
https://regex101.com/r/g4V0uw/1

(simple to modify the regexp (thanks to the one found in https://stackoverflow.com/questions/25425755/regular-expression-to-find-non-commented-java-lines see a test applicable to my current implementation with some simple twick (but necessarily which like line by line only half works...)
https://regex101.com/r/1kHY21/1
so as not to retain those which are in a "//" but not for those in a "/* */" on several lines with the implementation I made (reading line by line).
(There's also https://www.debuggex.com/ as an online tool for reg exp (I like state-rendered to understand what a regexp is doing...))

    Translator\s*\.\s*get\s*\(([^\\)]*)\)|(['"])(?:(?!\1|\\).|\\. )*\1|\/\/[^\n]*(?:\n|$)|\/\*(?:[^*]|\*(?!\/))*\*\/

Regular expression visualization

[Debuggex Demo](https://www.debuggex.com/r/9MioNPd3NTsa4t10)


Otherwise I had looked at the possibility of generating an issue ( https://github.com/marketplace/actions/create-issue-from-file created or completed with the content of a file if it exists ), I did tests on my fork with one of my branches https://github.com/PPAC37/Makelangelo-software/tree/searchInTheCode TraductionGetToFindTheKey_ci_create_issues but I haven't yet studied the trick not to create an issue, each time a build is done...

i-make-robots:
added missing strings to master.

coliss86:
> Really better as it is simpler than my #456
> 
> as you are basing yourself on my code, this also returns the Traduction.get in comment https://regex101.com/r/g4V0uw/1
> 
> (simple to modify the regexp (thanks to the one found in https://stackoverflow.com/questions/25425755/regular-expression-to-find-non-commented-java-lines see a test applicable to my current implementation with some simple twick (but necessarily which like line by line only half works...) https://regex101.com/r/1kHY21/1 so as not to retain those which are in a "//" but not for those in a "/* */" on several lines with the implementation I made (reading line by line). (There's also https://www.debuggex.com/ as an online tool for reg exp (I like state-rendered to understand what a regexp is doing... )) https://www.debuggex.com/r/xzJMkqryBQgBWc4h

Ok, I'll implement it asap

> Otherwise I had looked at the possibility of generating an issue ( https://github.com/marketplace/actions/create-issue-from-file created or completed with the content of a file if it exists ), I did tests on my fork with one of my branches https://github.com/PPAC37/Makelangelo-software/tree/searchInTheCodeTraductionGetToFindTheKey_ci_create_issues but I haven't yet studied the trick not to create an issue, each time a build is done...

I don't think it's necessary as this test will be executed of each PR and will fails if a key is missing, so you won't be able to merge it and thus you'll have to fix the code before.

coliss86:
@PPAC37 I have added a filter on lines starting with a comment.
There is no multi line comment for now in the code and I think it will be far more difficult to implement.

PPAC37:
@coliss86  Weel done ! 
And yes, it's not easy to find the right algorithm with the regexp that meets our needs. (I broke my teeth on it, especially since I'm quickly unable to express what I want in regexp (ex: when I want a negation like do not containe "//") and the only method I've found is to add a | "or part" to the regexp and to make a special case in the treatment of the match (if the group which interests me is null then it is the match on the " | or part" and therefore nothing to do for this match ...) )

src/test/java/com/marginallyclever/makelangelo/MissingTranslationTest.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Test
+    public void verifyThatMatcherWorks_results() throws IOException {
+        List<String> results = new ArrayList<>();
+
+        // matches.txt contains few entries with translation and one without. The list should not be empty
+        searchInAFile(new File("src/test/resources/translator/matches.txt"), results);
+
+        assertEquals(List.of("src/test/resources/translator/matches.txt:4: unknownKey"), results);
```
PPAC37, line 68: If you run it on a windows systeme ( file char separator "\\" vs "/" ) ... (this is one of the current CI test failed for this PR : 
~~~
Error:    MissingTranslationTest.verifyThatMatcherWorks_results:68 expected: <[src/test/resources/translator/matches.txt:4: unknownKey]> but was: <[src\test\resources\translator\matches.txt:4: unknownKey]>
~~~
)
maybe juste 
`"src/test/resources/translator/matches.txt:4: unknownKey".replaceAll("/", File.separator)` 
?
or else convert to URI the File for the output on the file ?

coliss86: thx, I modified with `.replaceAll("/", File.separator)`

src/test/java/com/marginallyclever/makelangelo/MissingTranslationTest.java
```diff
@@ -65,7 +65,7 @@ public void verifyThatMatcherWorks_results() throws IOException {
         // matches.txt contains few entries with translation and one without. The list should not be empty
         searchInAFile(new File("src/test/resources/translator/matches.txt"), results);
 
-        assertEquals(List.of("src/test/resources/translator/matches.txt:4: unknownKey"), results);
+        assertEquals(List.of("src/test/resources/translator/matches.txt:4: unknownKey".replaceAll("/", File.separator)), results);
```
PPAC37, line 68: Sorry i may have miss something
now we have
~~~
Error:  Errors: 
Error:    MissingTranslationTest.verifyThatMatcherWorks_results:68 � IllegalArgument cha...
~~~
and I'm confused as to why. encoding problem of my message if you copy pasted it?

coliss86: I changed the output of the test, on IntelliJ, I can now click in the log on the file and it opens directty to the file at the line. Also update the assertion to not fails on windows (hopefully)

image

PR #517, opened 2022-02-05
New gui translated elements

PPAC37:
a try to resolve some elements in https://github.com/MarginallyClever/Makelangelo-software/issues/513

I am not satisfied with what I propose for the enum.

https://github.com/MarginallyClever/Makelangelo-software/blob/c4b369c1713061acd0302a6fd4fdc881b40fe1d7/src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java#L14-L17

As I want to ensure that the CI test identifies the keys, I need a Translator.get("...") with simple fixed String.
But that makes the encoding and/or the enum unpleasant to look at from a code point of view.

Any ideas for finding a better compromise? or I drop the guarantee of the control of the translation keys (and the fact of explaining in the enum which element is used for the translation keys and therefore in the event of modification of the enum it will be necessary without remembering ...) ?

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
+++ b/src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
@@ -12,7 +12,7 @@ import com.marginallyclever.makelangelo.plotter.marlinSimulation.MarlinSimulatio
  * @author Dan Royer
  */
 public enum TurtleRenderFactory {
-	DEFAULT("Default", new DefaultTurtleRenderer(),"TurtleRenderFactory.DEFAULT"),// not coder friendly but allow later the CI test not to miss this translation keys.
+	DEFAULT("Default", new DefaultTurtleRenderer(),"TurtleRenderFactory.DEFAULT"),
 	BARBER_POLE("Barber pole", new BarberPoleTurtleRenderer(),"TurtleRenderFactory.BARBER_POLE"),
 	SEPARATE_LOOP("Separate loops",new SeparateLoopTurtleRenderer(),"TurtleRenderFactory.SEPARATE_LOOP"),
 	MARLIN_SIM("Marlin simulation",new MarlinSimulationVisualizer(),"TurtleRenderFactory.MARLIN_SIM");
--- a/src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
+++ b/src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
@@ -12,16 +12,14 @@ import com.marginallyclever.makelangelo.plotter.marlinSimulation.MarlinSimulatio
  * @author Dan Royer
  */
 public enum TurtleRenderFactory {
-	DEFAULT("Default", new DefaultTurtleRenderer(),Translator.get("TurtleRenderFactory.DEFAULT")),// not coder friendly but allow later the CI test not to miss this translation keys.
-	BARBER_POLE("Barber pole", new BarberPoleTurtleRenderer(),Translator.get("TurtleRenderFactory.BARBER_POLE")),
-	SEPARATE_LOOP("Separate loops",new SeparateLoopTurtleRenderer(),Translator.get("TurtleRenderFactory.SEPARATE_LOOP")),
-	MARLIN_SIM("Marlin simulation",new MarlinSimulationVisualizer(),Translator.get("TurtleRenderFactory.MARLIN_SIM"));
+	DEFAULT("Default", new DefaultTurtleRenderer()),
+	BARBER_POLE("Barber pole", new BarberPoleTurtleRenderer()),
+	SEPARATE_LOOP("Separate loops",new SeparateLoopTurtleRenderer()),
+	MARLIN_SIM("Marlin simulation",new MarlinSimulationVisualizer());
 
 	private final TurtleRenderer turtleRenderer;
 
 	private final String name;
-	
-	private final String translatedText;
 
 	/**
 	 * 
@@ -29,10 +27,9 @@ public enum TurtleRenderFactory {
 	 * @param turtleRenderer
 	 * @param translatedText the text (translated) N.B. : do not rely on the name (used in findByName) to avoid a bad translation (which can give for two elements the same value) to assert any translation may not create issues... 
 	 */
-	TurtleRenderFactory(String name, TurtleRenderer turtleRenderer, String translatedText) {
+	TurtleRenderFactory(String name, TurtleRenderer turtleRenderer) {
 		this.name = name;
 		this.turtleRenderer = turtleRenderer;
-		this.translatedText = translatedText;
 	}
 
 	public TurtleRenderer getTurtleRenderer() {
@@ -43,14 +40,6 @@ public enum TurtleRenderFactory {
 		return name;
 	}
 	
-	/**
-	 * 
-	 * @return The text (translated) to used as text in the GUI ...
-	 */
-	public String getTranslatedText() {
-		return translatedText;
-	}
-
 	public static TurtleRenderFactory findByName(String name) {
 		return Arrays.stream(values())
 				.filter(enumValue -> enumValue.getName().contains(name))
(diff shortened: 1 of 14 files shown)


src/main/resources/languages/french.xml
```diff
@@ -2763,4 +2763,45 @@
...
+
+    
+    <string>
+        <key>TurtleRenderFactory.DEFAULT</key>
+        <value>Défaut</value>
+        <hint></hint>
+    </string>
+    <string>
+        <key>TurtleRenderFactory.BARBER_POLE</key>
+        <value>Poteau de barbier</value>
```
coliss86, line 2785: The traduction is funny ;-)

src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
```diff
@@ -11,18 +12,27 @@
...
-
-	TurtleRenderFactory(String name, TurtleRenderer turtleRenderer) {
+	
+	private final String traductedText;
+
+	/**
+	 * 
+	 * @param name
+	 * @param turtleRenderer
+	 * @param traductedText the text (traducted) N.B. : not base on the name (used in findByName) to avoid a bad traduction (that give for two elements the same value) to assert any traduction may not create issues...
```
coliss86, line 30: the word "traduction" does not exist in english, you should use "translation" instead

coliss86 (changes requested):
Very nice work, few things to adjust

i-make-robots:
Well… previously it was a factory instead of an enumeration. Idk if that makes the presentation any better.

coliss86:
> Well… previously it was a factory instead of an enumeration. Idk if that makes the presentation any better.

It is an enum as previsously, the type did not change. I don't understand your point.

src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
PPAC37: Thanks ! and my apologies for my poor level in English.

src/main/resources/languages/french.xml
PPAC37: I'm struggling to find a translation that makes sense as it's a cultural idiom...
I will try something else ...

PPAC37:
> It is an enum as previsously, the type did not change. I don't understand your point.

@coliss86  I think @i-make-robots is talking about an adaptation prior to my edits/PR.

PPAC37:
I forgot to put a file in a commit (which broke the build) 
then I recently made a commit that fixes that, but strangely it doesn't appear here.

I don't know if I made a bad manipulation, or if github takes its time or if it's normal after a failed build for a PR...
Edit : the commit https://github.com/PPAC37/Makelangelo-software/commit/3dd8cf5fda6600036dec0f40aa0a2feb82a78f38

src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
```diff
@@ -11,18 +12,27 @@
...
  */
 public enum TurtleRenderFactory {
-	DEFAULT("Default", new DefaultTurtleRenderer()),
-	BARBER_POLE("Barber pole", new BarberPoleTurtleRenderer()),
-	SEPARATE_LOOP("Separate loops",new SeparateLoopTurtleRenderer()),
-	MARLIN_SIM("Marlin simulation",new MarlinSimulationVisualizer());
+	DEFAULT("Default", new DefaultTurtleRenderer(),Translator.get("TurtleRenderFactory.DEFAULT")),// not coder friendly but allow later the CI test not to miss this translation keys.
+	BARBER_POLE("Barber pole", new BarberPoleTurtleRenderer(),Translator.get("TurtleRenderFactory.BARBER_POLE")),
+	SEPARATE_LOOP("Separate loops",new SeparateLoopTurtleRenderer(),Translator.get("TurtleRenderFactory.SEPARATE_LOOP")),
+	MARLIN_SIM("Marlin simulation",new MarlinSimulationVisualizer(),Translator.get("TurtleRenderFactory.MARLIN_SIM"));
```
i-make-robots, line 18: When does this run?  I mean, how do we know it runs after Translator has loaded a language?

i-make-robots: how do we *guarantee*

coliss86: no it won't, you are right the correct language may not be loaded before the enum is loaded :'(

PPAC37: So my apologies for this proposal which does not guarantee the correct translation.

As I'm out of ideas, and in order to keep things simple in the code, I'm not going to use a Translator.get("...") expression recognized by the IC test for missing translation keys.

Therefore, for the translation of the items of this enumeration in the menu, if translation keys are missing, will not be detected by the CI test.

i-make-robots: Enum sounds good to make sure the classes exist at compile time.  A factory and a `testMakeOneOfEach()` would do the same job, plus the factory will definitely run after initialization and use less ram/startup time if never called at all.

coliss86: What you can do is simply put the Key in the enum and when the translation in needed, call the translator with the key, like : 
```
DEFAULT("Default", new DefaultTurtleRenderer(), "TurtleRenderFactory.DEFAULT")
```

If the CI test is a worry, don't worry :) The regexp will on catch `Translator.get("XXX")`, a pattern with double quote, so it won't hurt.

PPAC37: Thanks, much more logical indeed to keep the translation keys in the enum

src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFactory.java
```diff
@@ -32,6 +42,14 @@ public TurtleRenderer getTurtleRenderer() {
 	public String getName() {
 		return name;
 	}
+	
+	/**
+	 * 
+	 * @return The text (translated) to used as text in the GUI ...
```
coliss86, line 48:
```suggestion
	 * @return The translated text to be used as text in the GUI
```

src/main/resources/languages/french.xml
```diff
@@ -2763,4 +2763,50 @@
...
+    </string>
+    <string>
+        <key>TurtleRenderFactory.MARLIN_SIM</key>
+        <value>Simulation vitesses marlin</value>
+        <hint></hint>
+    </string>
+    <string>
+        <key>TurtleGenerators.LearnMore.Link.Text</key>
+        <value>En savoir plus</value>
+        <hint>Used in many Generator Panel as link text to Wikipedia URLs</hint>
```
coliss86, line 2801:
```suggestion
        <hint>Utilisé dans différents panel comme texte de lien vers Wikipedia</hint>
```

i-make-robots:
Is this ready now?

PPAC37:
> Is this ready now?

@i-make-robots For me, yes.
But maybe @coliss86 will have other ideas of things to improve.

coliss86:
> @i-make-robots For me, yes. But maybe @coliss86 will have other ideas of things to improve.

Nothing from me, I already made a review and everything has been done 💪

i-make-robots:
Uh... after merging this english is no longer a language option, even though english.xml still exists.

PPAC37:
> Uh... after merging this english is no longer a language option even though english.xml still exists.
> 
> edit: ....and now it's back. wtf.

Maybe an on-the-fly compilation story that wasn't fully realized...

As the .xml language files are in the resources, they are not necessarily copied (unless you force a build). And you may end up with a version of english.xml which had a `<value><\value>` (empty)

 and if cumulated with a target folder based on a compilation of the old version of TraductionLanguage (which crashes the load of a language if there is an (empty) `<value><\value>` ...

Or maybe I messed up my edits to accept `<value><\value>` (empty).

If you have this problem again you will have to question what I did in https://github.com/MarginallyClever/Makelangelo-software/pull/535

PR #521, opened 2022-02-05
💄 beginning of the baud rate configuration

coliss86:
New settings: 
image

Damn it, it is not at the correct place. It should be in the connection panel as said in https://github.com/MarginallyClever/Makelangelo-software/issues/482#issuecomment-1025264946


coliss86:
Abandoned, will do a new one with the drop down at the correct place
The translation part is added in a new PR #526

PR #522, opened 2022-02-06
refactoring utility function in SelectReadOnlyText to create JeditoPane whith HyperLinkListenerAndToolTips

PPAC37:
Surely a little more greedy but seems much more pleasant / generic and user friendly. 
image
image

what do you think?

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
+++ b/src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
@@ -9,7 +9,6 @@ import javax.swing.JEditorPane;
 
 import javax.swing.ToolTipManager;
 import javax.swing.event.HyperlinkEvent;
-import javax.swing.event.HyperlinkListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
+++ b/src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
@@ -2,56 +2,83 @@ package com.marginallyclever.makelangelo.select;
 
 import java.awt.BorderLayout;
 import java.awt.Desktop;
-import java.awt.event.MouseAdapter;
-import java.awt.event.MouseEvent;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import javax.swing.JEditorPane;
 
-import javax.swing.JLabel;
+import javax.swing.ToolTipManager;
+import javax.swing.event.HyperlinkEvent;
+import javax.swing.event.HyperlinkListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
- * Read only HTML-rich text that can display AT MOST one clickable link.
- * @author Dan Royer
+ * Read only JEditorPane with an HyperLinckListener for text/html contening or not multiple html link.
+ * With clickable link (Desktop.BROWSE) and ToolTips when hovered over .
+ * 
+ * @author Dan Royer (and modified by PPAC37)
  * @since 7.24.0
  */
 public class SelectReadOnlyText extends Select {
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -8918068053490064344L;
-	private static final String A_HREF = "<a href=";
-	private static final String HREF_CLOSED = ">";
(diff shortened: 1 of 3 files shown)


src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
```diff
@@ -2,56 +2,83 @@
...
+ * 
+ * @author Dan Royer (and modified by PPAC37)
  * @since 7.24.0
  */
 public class SelectReadOnlyText extends Select {
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -8918068053490064344L;
-	private static final String A_HREF = "<a href=";
```
i-make-robots, line 23: will break if someone does something dumb like "<a     href    ="

src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
```diff
@@ -2,56 +2,83 @@
...
-	private static final String HREF_CLOSED = ">";
+	private static final long serialVersionUID = -8918068053490064344L;// TODO to regenerat ?
```
i-make-robots, line 27: this is never actually serialized, so I don't believe it needs to be regenerated.

src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
```diff
@@ -2,56 +2,83 @@
...
+	 * @param sToSetAsTextToTheHtmlEditorPane
+	 * @return 
+	 */
+	public static JEditorPane createJEditorPaneWithHyperlinkListenerAndToolTipsForDesktopBrowse(String sToSetAsTextToTheHtmlEditorPane) {
+		JEditorPane createdJEditorPane =  new JEditorPane();
+		createdJEditorPane.setEditable(false);
+		createdJEditorPane.setOpaque(false);
+		createdJEditorPane.setContentType("text/html");
+		createdJEditorPane.setText(sToSetAsTextToTheHtmlEditorPane);
+		createdJEditorPane.addHyperlinkListener(new HyperlinkListener() {
```
i-make-robots, line 50:
```
createdJEditorPane.addHyperlinkListener( (hyperlinkEvent)->{
```
maybe?

src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
```diff
@@ -2,56 +2,83 @@
...
+		jEdPane = createJEditorPaneWithHyperlinkListenerAndToolTipsForDesktopBrowse("<html>"+ labelKey+"</html>");
+		this.add(jEdPane,BorderLayout.CENTER);
+	}
+	
+	/**
+	 * Create a JEditorPane not editable for text/html contente, with an HyperLinkListener to Desktop Browse (when clicked) and show a ToolTips with the URL hovered.
+	 * @param sToSetAsTextToTheHtmlEditorPane
+	 * @return 
+	 */
+	public static JEditorPane createJEditorPaneWithHyperlinkListenerAndToolTipsForDesktopBrowse(String sToSetAsTextToTheHtmlEditorPane) {
```
i-make-robots, line 44: I love this. never be afraid to use long and very descriptive names.

src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
PPAC37: I hadn't paid attention, I guess with the JEditorPane and HyperlinkListener this should no longer be a problem.

src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
PPAC37: :) At the time I said to myself that I may be abusing ... but after the video on how to code well, I adhere to "the code must explain itself"

src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
PPAC37: Thanks, comment removed.

src/main/java/com/marginallyclever/makelangelo/select/SelectReadOnlyText.java
PPAC37: Thank you I too often forget the lambda expressions. I have to revisit my habits.

PR #523, opened 2022-02-06
LoadGCode.java a draft to read back color from a "M0 Ready ..." color change.

PPAC37:
Try saving a generated graph paper as .gcode and load the saved .gcode. (currently we lose the color change.)

This will read and apply the color change normally from the .gcode file.

But it's a rough draft because I have a bit of a mess. (int-valued color name in a bidirectional map or not, w3c color naming convention, ...)


PPAC37:
As I have no positive or negative feedback on this proposal, 

I tell myself that it should not be of great interest 
so I close this request.

PR #525, opened 2022-02-06
🚚 Rename NetworkSession* to Connection*

coliss86:
The word Network usually refers to TCP communication, but here we want something independent as described in TransportLayer

### DIFF
--- a/src/main/java/com/marginallyclever/communications/Communication.java
+++ b/src/main/java/com/marginallyclever/communications/Communication.java
@@ -9,7 +9,7 @@ import java.util.List;
  * @author Peter Colapietro
  * @since v7
  */
-public abstract class NetworkSession {
+public abstract class Communication {
 
 	private String name = "";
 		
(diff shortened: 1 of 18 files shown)


i-make-robots:
It is not a connection.  In the OSI model (https://en.wikipedia.org/wiki/OSI_model) at this level of abstraction it is in the Session layer.

i-make-robots:
I'm pretty sure each class mentions what level of the OSI layer it belongs in as part of the class header javadoc.

coliss86:
All the code in `com.marginallyclever.communications` is about the communication of the robot regardless the protocol and the transport, isn't it ?

Maybe `Connection` is not a correct word, but neither is `Network`. In the page you gave, especially at the table [Comparison to other networking suites](https://en.wikipedia.org/wiki/OSI_model#Comparison_to_other_networking_suites) serial transport is at the physical layer and TCP/IP is at the session.

> I'm pretty sure each class mentions what level of the OSI layer it belongs in as part of the class header javadoc.

Unfortunately, the javadoc of the classes of this PR does not mention any layer and the [`src/main/java/com/marginallyclever/communications/package-info.java`](https://github.com/MarginallyClever/Makelangelo-software/blob/master/src/main/java/com/marginallyclever/communications/package-info.java#L2) mentions the word `Connection`.

PR #526, opened 2022-02-06
🔊 add log when a translation is missing

coliss86:


i-make-robots:
I thought we were already through CI errors when translations are missing.  why, then, do we also need a log?

PPAC37:
> I thought we were already through CI errors when translations are missing. why, then, do we also need a log?

In the case of a personal translation file of the user (possibly a translator) placed next to the .jar.

coliss86:
It is useful when developing but you are right, the CI is already there.

PR #528, opened 2022-02-06
💄 add a button with multiple sets of icons, texts and optionally foreground colors

coliss86:
It replaces the `ButtonIcon bConnect` of `ChooseConnection` with the same functionality and allow this new button to be re-used elsewhere.

### DIFF
--- a/src/main/java/com/marginallyclever/convenience/ToggleButtonIcon.java
+++ b/src/main/java/com/marginallyclever/convenience/ToggleButtonIcon.java
@@ -1,0 +1,55 @@
+package com.marginallyclever.convenience;
+
+import com.marginallyclever.makelangelo.Translator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.swing.*;
+import java.awt.*;
+import java.util.List;
+import java.util.stream.Collectors;
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/marginallyclever/convenience/ToggleButtonIcon.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * A button with multiple sets of icons, texts and optionally foreground colors
+ */
+public class ToggleButtonIcon extends JButton {
+    private static final Logger logger = LoggerFactory.getLogger(ToggleButtonIcon.class);
+
+    private final List<String> translations;
+    private final List<ImageIcon> icons;
+    private final List<Color> foregroundColors;
```
i-make-robots, line 20: are these three lists always the same length?

coliss86: yes. `foregroundColors` is optionnal

i-make-robots: if they were all the same length I would say make a single list of a new subclass.  if foreground color is optional then don't bother.

PR #530, opened 2022-02-08
Work on bug 513

i-make-robots:
@PPAC37 does this make things better?

PR #533, opened 2022-02-09
Work on bug 513 xml validation fix and allowing empty value for a translation

PPAC37:
This adding to #530 modification, normaly allow #530 to pass CI test to allow empty value for translation in the xml validation (xsd shema with to hight restriction for the contente)  and in the load (TranslatorLanguage.java) of a language file.

### DIFF
--- a/src/test/resources/translator/language_no_dup_key.xsd
+++ b/src/test/resources/translator/language_no_dup_key.xsd
@@ -21,7 +21,7 @@
                     <xs:complexType>
                         <xs:sequence>
                             <xs:element type="stringNotEmpty" name="key"/>
-                            <xs:element type="stringNotEmpty" name="value"/>
+                            <xs:element type="xs:string" name="value"/><!-- type from "stringNotEmpty" changed to "xs:string" to remove restriction to allow empty value -->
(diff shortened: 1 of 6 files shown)


src/test/resources/translator/language_no_dup_key.xsd
```diff
@@ -21,7 +21,7 @@
                     <xs:complexType>
                         <xs:sequence>
                             <xs:element type="stringNotEmpty" name="key"/>
-                            <xs:element type="stringNotEmpty" name="value"/>
+                            <xs:element type="xs:string" name="value"/><!-- type from "stringNotEmpty" changed to "xs:string" to remove restriction to allow empty value -->
```
coliss86, line 24: I don't think the comment is necessary, it just described what is in the code, not why.

coliss86:
There 2 things in this PR : something about the center cartesian button which does nothing anymore and allowing empty value in the translation. 
The title of the PR (which will be the final commit) should reflect what is really done.

In my opinion, there should be 2 PR: one for the translation part, and a second on the home button.
One commit is on a subject only. Easy to say, hard to do, but it is case, easy :)

PR #540, opened 2022-02-12
🔊 move log to user home and add dropdown menu to the log panel to select the log file to view

coliss86:
Here is the enhanced log panel: 
image

@PPAC37 @i-make-robots can you please test this feature? It should solved the problem that @PPAC37 found in the review of https://github.com/MarginallyClever/Makelangelo-software/pull/536#issuecomment-1037228202: 
> In fact this is a security issues. As my user have no admin right the path defined in the shortcut is not writable.


PPAC37:
Ok for me (Ubuntu 18.04, run from NetBeans and from `java -jar ...` )
image

coliss86:
> Ok for me (Ubuntu 18.04, run from NetBeans and from `java -jar ...` ) 

perfect, thanks !

i-make-robots:
Does it cycle log files after 5 or 10?  users won't want to be confused with many log files.

coliss86:
it keeps 1 old files. It is specified here: https://github.com/MarginallyClever/Makelangelo-software/blob/master/src/main/resources/logback.xml#L25

PR #542, opened 2022-02-13
some french.xml translation corrections/better choice

PPAC37:


src/main/resources/languages/french.xml
```diff
@@ -296,7 +296,7 @@
     </string>
     <string>
         <key>SierpinskiTriangleName</key>
-        <value>Fractale du triangle de Sierpinski</value>
+        <value>Triangle de Sierpiński</value>
         <hint>[name of generator]</hint>
         <!--	"SierpinskiTriangleName"
```
coliss86, line 301: what is the purpose of those comments ? it will be outdated very shortly when the code is updating.
As the `<key>` is almost unique in the code, it's trivial to find where it is used by doing a search, isn't it ?

src/main/resources/languages/french.xml
```diff
@@ -3,7 +3,7 @@
     <meta>
         <name>Français (fr_FR)</name>
         <author>https://github.com/PPAC37</author>
-        <!-- draft v0.4.1 04/02/2022 -->
+        <!-- draft v0.4.2 13/02/2022 -->
```
coliss86, line 6: It is not a draft anymore ! the date and the version are also not useful, the history of git serves this purpose.

src/main/resources/languages/french.xml
PPAC37: I haven't deleted them for the days when I would like to work on renaming key to correspond to https://github.com/MarginallyClever/Makelangelo-software/discussions/392#discussion-3761569 
>  if a translated string is only used in one class it should be ClassName.KeyName

... to have an idea of the extent of the work to be done 
... as normally I have something to regenerate them automatically, I will remove them.

src/main/resources/languages/french.xml
PPAC37: old habits ... with git, indeed, it is no longer useful.

coliss86 (approved):
Perfect !

coliss86:
there is a typo in the title of the PR : "choi_c_e' instead of "choise"

PR #546, opened 2022-02-14
💄 add the full git revision in the about dialog and in logs

coliss86:
Here is the about dialog box:
image

The version is "hash-ee8c91a-dirty": 
- `hash` is the name of my branch
- `ee8c91a` is the hash commit
- `dirty` is the state of the working copy if some modifications has been made to the code base or not.

Fix https://github.com/MarginallyClever/Makelangelo-software/issues/543


PPAC37:
I love ! Good work !

PR #550, opened 2022-02-16
Fix for seahorse 

PPAC37:
as explaine in https://github.com/MarginallyClever/Makelangelo-software/issues/353#issuecomment-1039446478
avoid having some drawn lines which should be displacements.

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
+++ b/src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
@@ -367,7 +367,7 @@ public class LoadSVG implements TurtleLoader {
 		ArrayList<Point2D> points = b.generateCurvePoints(0.1);
 		for(Point2D p : points) myTurtle.moveTo(p.x,p.y);
 		pathPoint.set(p3);
-		isNewPath=false;
+		isNewPath=true;
 	}
 
 	private void doLineToRel(SVGPathSeg item, Matrix3d m) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
```diff
@@ -367,7 +367,7 @@ private void doCubicCurveAbs(SVGPathSeg item, Matrix3d m) {
 		ArrayList<Point2D> points = b.generateCurvePoints(0.1);
```
coliss86, line 367:
```suggestion
		List<Point2D> points = b.generateCurvePoints(0.1);
```

i-make-robots: why is it better to use generic List?  One extra include....

coliss86: you don't need the method specific to an ArrayList in this part of the code, you just need a List which is the interface.

i-make-robots: I don't need an extra include, either!  If I had a factory that was returning a List then sure, generic allows for more flex.  But here it's of no benefit.

coliss86: `import` are generated by the IDE, so it is not an issue. your main problem is why ArrayList over List. I may not using the right word to explain so, instead, I suggest you to read this small answer: https://stackoverflow.com/questions/147468/why-should-the-interface-for-a-java-class-be-preferred

PR #551, opened 2022-02-16
More resilient LoadScratch3 with some fix (orientation)

PPAC37:
* [x] see https://github.com/MarginallyClever/Makelangelo-software/issues/534#issuecomment-1037317682
* [ ] In some case the Turtle setColor trouble the render result ... have to investigate ... Edit : ResizeTurtleToPaperAction resize = new ResizeTurtleToPaperAction(myPaper,false,""); seem to have trouble with colorChange in some case... cause if commented no issues until manualy activate it ...
* [x] hack loop forever allowed but iterated only once.
* [x] No error on loading a not clean Scratch project ( a variable block left in a script ...)
* [x] can render multiple block opcode event_whenflagclicked

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadScratch3.java
+++ b/src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadScratch3.java
@@ -34,6 +34,8 @@ public class LoadScratch3 implements TurtleLoader {
 	private static final Logger logger = LoggerFactory.getLogger(LoadScratch3.class);
 	private final String PROJECT_JSON = "project.json";
 	
+	boolean verboseParse = false;
+	
 	private class Scratch3Variable implements Cloneable {
 		public String name;
 		
@@ -46,6 +48,7 @@ public class LoadScratch3 implements TurtleLoader {
 			this.value=defaultValue;
 		}
 		
+		@Override
 		public String toString() {
 			return //uniqueID+" "+
 					name+"="+value;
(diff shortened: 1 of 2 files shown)


PPAC37:
> probably in doStart() the Turtle should be set to -90. that way if the first comand is "go forward" it moves in the correct direction.

it seems that at the initialization of a project the orientation is at 90° under scratch 
https://scratch.mit.edu/projects/645552141
(therefore should be at 0° in turtle doStart() ).

(but not if you do several times the green flag ... there it starts again on the value at the end of the previous exec ... it must be a technique to make you think about doing an initialization phase in your project scratch ... like adding a Go to (x=0, y=0) ...

i-make-robots:
on green flag should not happen more than once in any script.  I believe the Scratch app won't let you have more than one green flag.

PPAC37:
I expressed myself badly, I was talking about clicking several times on the green flag under scratch.
But your remark made me do a test. 
And unfortunately for us, it seems possible to have several times the "when green flag clicked" in the same script.

https://scratch.mit.edu/projects/645649727
image

i-make-robots:
You're right.  Each one would run separately, not in parallel.  We can't wipe the Turtle to nothing on green flag.

i-make-robots:
that suggests the problem is in resize, not in LoadScratch3

PPAC37:
I think I found and solved my rendering bug story for some color changes in a Scratch3 project.

i-make-robots:
I think your Turtle fix is correct.

PR #553, opened 2022-02-16
🤖 reset the robot on connect

coliss86:
When opening a connection to the robot, it is now reset.
@i-make-robots can you check if it also works on your robot please as I don't have the same hardware as you. 

Taken from https://github.com/arduino/Arduino/blob/18f55445ab95b193671306688f2a8f4877e95e17/arduino-core/src/processing/app/Serial.java#L103 and https://reprap.org/forum/read.php?4,888289,888290#msg-888290


i-make-robots:
The robot has always reset when i open a connection.  your hardware is behaving differently from RUMBA boards.  I will check it as soon as I can.

PR #554, opened 2022-02-16
💄 change the shortcut modifier to COMMAND if macOS is detected

coliss86:
on macOS, shortcut are not with the key CTRL but with the COMMAND key. Here is an enhancement that take care of this.

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
+++ b/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
@@ -127,6 +127,10 @@ public final class Makelangelo {
 	private RecentFiles recentFiles;
 	private OpenFileChooser openFileChooser;
 
+	private static int SHORTCUT_CTRL = InputEvent.CTRL_DOWN_MASK;
+	private static int SHORTCUT_ALT = InputEvent.ALT_DOWN_MASK;
+
+	
 	// drag files into the app with {@link DropTarget}
 	@SuppressWarnings("unused")
 	private DropTarget dropTarget;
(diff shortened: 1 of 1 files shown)


src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -547,14 +553,14 @@ private JMenu createFileMenu() {
...
-		buttonSaveFile.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_S, InputEvent.CTRL_DOWN_MASK));//"ctrl S"
+		buttonSaveFile.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_S, SHORTCUT_CTRL));//"ctrl S"
 		menu.add(buttonSaveFile);
 
 		menu.addSeparator();
 				
 		JMenuItem buttonAdjustPreferences = new JMenuItem(Translator.get("MenuPreferences"));
 		buttonAdjustPreferences.addActionListener((e)-> myPreferencesPanel.run(mainFrame));
-		buttonAdjustPreferences.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_P, InputEvent.ALT_DOWN_MASK));//"alt P"
+		buttonAdjustPreferences.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_P, SHORTCUT_ALT));//"alt P"
```
i-make-robots, line 563: will this conflict with command+P or ctrl+p (print)?  This is the only instance of SHORTCUT_ALT.  Maybe a shortcut that doesn't require ALT makes life easier?

coliss86: you are right, on a mac, preferences pane are mapped to COMMAND + ,
I will change this

coliss86: I made the modification

PR #563, opened 2022-02-19
✅ add unit tests for FileAccess and LoadSVG

coliss86:


PPAC37:
Perhaps this one 
https://github.com/MarginallyClever/Makelangelo-software/blob/02b661cb84fb2f73f3412e697637457aca172ad3/src/test/resources/svg/multi_shapes1.svg
is not to be tested immediately because it is not correctly interpreted?

image

coliss86:
The blue / violet parts are not interpreted by the app, infill are skipped I think.
Eclipse is Ok, I see the same thing :-)

coliss86:
Oh, you mean the triangle ? I didn't notice it, good catch ! I'll move it out the way

PPAC37:
> Oh, you mean the triangle ? I didn't notice it, good catch ! I'll move it out the way

Yes the ellipse which makes a triangle.
(Phew for a while you made me switch to a questioning of my mental state. (it's good then I'm not too crazy yet :) ))
Maybe we should have a special folder for test files that we don't pass yet.

coliss86:
I have created an issue with this one: #568

PR #565, opened 2022-02-20
🔥 remove module-info

coliss86:
@PPAC37 @i-make-robots please test on your setup if we can get ride of this file !


PPAC37:
I have already tried to delete the module-info.java in some of my branches.

and apart from a little moment when my NetBeans IDE is destabilized (puts a lot of stuff in error, you have to restart it or let it work or else force a clean and build so that everything is in order)

So ok for me.

i-make-robots:
Seems Ok for me, too.

PR #567, opened 2022-02-20
✅ add unit tests for LoadScratch3

coliss86:
This PR is based on the work on #563 and show duplicates code between the both, as soon as #563 is merged, I'll update #567


PPAC37:
@coliss86  As I bang my head on the wall with my history of Scratch3 and orientation (ha sweet trigo ...) I'm not sure to see if there are default/subtleties in your test.

Reading the code for the LoadScratch3 Test makes me say : yes perfect !. 

(Just maybe make available the utility function myTurtle to TextFile (I haven't seen it... but maybe I'm temporarily blind to banging my head on the walls ;-) ) to facilitate the creation / modification of the expected files)

coliss86:
there is no default or anything in those tests :-) their purpose is to ensure there is no regression in the future. So whatever you'll do fix the issue, at the end, the file you used and the expected behavior has to be added.

PPAC37:
In fact it seems to me that a Test is missing for Turtle
Which checks that a turtle.history.toString() is stable (that TurtleMove.toString is stable...)

But we can say that it is implicitly done ( If All Load* test fails it's probably turtle.history.toString() ...)

coliss86:
Sure, those tests are just a start, feel free to add some more or, tell me what you think that could be added to check small things. 
Loading a whole file as I have done, is more an integration test rather than a unit test that verify one behavior.

src/test/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVGTest.java
```diff
@@ -0,0 +1,90 @@
...
+        // given
+        TurtleLoader loader = new LoadSVG();
+
+        // when
+        Turtle turtle = loader.load(LoadSVGTest.class.getResourceAsStream(svgFile));
+
+        // then
+        assertNotNull(turtle);
+        assertNotNull(turtle.history);
+        assertEquals(readFile(expectedFile), turtle.history.toString());
```
i-make-robots, line 88: If we change the quality of ellipse or bezier interpolation the related tests will break.  That's a tomorrow problem.

coliss86: yes, it is normal to update test when the underlying is modified.

Issue #568, opened 2022-02-20
SVG ellipse not interpreted correctly


coliss86:
**Steps to make the bug happen**
1. Load the attached svg [ellipse.svg](https://user-images.githubusercontent.com/23615562/154847077-b1173050-e782-4df0-b16b-344cf3de5328.svg)
2. the app convert it to a triangle : 
![image](https://user-images.githubusercontent.com/23615562/154847180-53c5b9b2-7797-4fd1-92a8-5d3bf4ec718e.png)
 

**What was supposed to happen**
It is a ellipse
![image](https://user-images.githubusercontent.com/23615562/154847168-7b481155-6db9-4689-9c18-b07fdbbef14b.png)

**Platform (please complete the following information):**
 - Version: master
 - OS: macOS

**Log file**
```
2022-02-20 15:13:09,886 DEBUG c.m.makelangelo.Translator - starting translator... 
2022-02-20 15:13:09,900 DEBUG c.m.makelangelo.Translator - Default language = English 
2022-02-20 15:13:09,984 DEBUG c.m.makelangelo.Translator - loadConfig: User Preference Node: /DrawBot/Language=English 
2022-02-20 15:13:10,465 DEBUG c.m.makelangelo.paper.Paper - loadConfig() before Paper Width = 420.00 Height = 594.00 (left=-210.00 right=210.00 top=297.00 bottom=-297.00) centerShift(X= 0.00 Y= 0.00) color (255,255,255) 
2022-02-20 15:13:10,467 DEBUG c.m.makelangelo.paper.Paper - loadConfig() after Paper Width = 210.00 Height = 297.00 (left=-105.00 right=105.00 top=148.50 bottom=-148.50) centerShift(X= 0.00 Y= 0.00) color (255,255,255) 
2022-02-20 15:13:11,698 DEBUG c.m.util.PropertiesFileHelper - version: 7.31.0 
2022-02-20 15:13:11,699 DEBUG c.m.util.PropertiesFileHelper - fullGitRevision: dirty 
2022-02-20 15:13:11,726 DEBUG c.m.makelangelo.RecentFiles - loading recent files 
2022-02-20 15:13:13,088 DEBUG c.m.m.preview.PreviewPanel -   get GL capabilities... 
2022-02-20 15:13:13,092 DEBUG c.m.m.preview.PreviewPanel -   starting animator... 
[...]

PR #572, opened 2022-02-20
added ellipse perimeter approximation (for #568)

i-make-robots:


src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
```diff
@@ -367,7 +367,7 @@ private void doCubicCurveAbs(SVGPathSeg item, Matrix3d m) {
 		List<Point2D> points = b.generateCurvePoints(0.1);
 		for(Point2D p : points) myTurtle.moveTo(p.x,p.y);
 		pathPoint.set(p3);
-		isNewPath=true;
+		isNewPath=false;
```
PPAC37, line 370: My fix for the seahorses created problems elsewhere and you intentionally corrected it, where is it that you started from a version that did not have this fix?

i-make-robots: I started from node-based-editor, realized I was in the wrong branch, copied the whole file, went to master, made a new branch, pasted, and that's how it got lost.  I didn't want to refactor the whole thing and I didn't mean to step on your change.  Thank goodness for git and for you!  I will change it right now.

PPAC37:
No problem, i mix quickly when working on several branches too.

I started to use the export in .diff file of a commit or of uncommitted changes of a local modified version to then apply the diff file on the branch that I want rather than copying a file... 
seems that there is simpler but it allows me to keep my modifications and to be able to reapply them to the needs on the branch that I want.

coliss86:
Can you first merge https://github.com/MarginallyClever/Makelangelo-software/pull/563 and then add a new unit test in LoadSVGTest with the file given in the issue to ensure it works as intended please ?

src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
```diff
@@ -239,112 +240,111 @@ private void arcTurtle(Turtle turtle,double cx,double cy,double rx,double ry,dou
...
-				double c=rx*Math.cos(v);
-				v2 = transform(cx+c,cy+s,m);
-				myTurtle.moveTo(v2.x,v2.y);
-			}
-			v2 = transform(cx+rx,cy,m);
+
+			double perimeterOfAnEllipseApprox = Math.PI * 2.0 * Math.sqrt((ry*ry + rx*rx)/2.0);
+			double steps = Math.max(3,perimeterOfAnEllipseApprox);
+			steps = Math.min(60,steps);
+			printEllipse(m, cx, cy, rx, ry, perimeterOfAnEllipseApprox);
```
PPAC37, line 293: un-used "steps" variable ?

src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
```diff
@@ -239,112 +240,111 @@ private void arcTurtle(Turtle turtle,double cx,double cy,double rx,double ry,dou
...
+	private void printEllipse(Matrix3d m, double cx, double cy, double rx, double ry, double perimeterOfAnEllipseApprox) {
+		Vector3d v2;
+		for(double i = 1; i<perimeterOfAnEllipseApprox; ++i) {
+			double v = (Math.PI*2.0) * (i/perimeterOfAnEllipseApprox);
+			double s=ry*Math.sin(v);
+			double c=rx*Math.cos(v);
+			v2 = transform(cx+c,cy+s,m);
 			myTurtle.moveTo(v2.x,v2.y);
-	    }
+		}
```
PPAC37, line 305: Just to be sure to understand it right (sorry long discussion)

m is a transformation matrix (x,y translation and rotation (no scale))

we draw here small segments of 1 unit (i++) (assuming that the unit/granule is 1mm (before any other turtle transformation))

whatever size our circle/ellipse is, we should have a "segment" resolution/granul of ~1?

otherwise, the segments are drawn inside the arc (only their endpoints fall exactly on the path of the arc)?

We are in the case circle (green) our aproximation is the bleu ?
image



from https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Primitive_Solids#sphere 

because circular paths are approximated with polygons inscribed within in a circle. The points of the polygon are on the circle, but straight lines between are inside. To have all of the hole larger than the true circle, the polygon must lie wholly outside of the circle (circumscribed). [Modules for circumscribed holes](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/undersized_circular_objects)

src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
i-make-robots: whoops

src/main/java/com/marginallyclever/makelangelo/makeArt/io/vector/LoadSVG.java
i-make-robots: Yes, circles become slightly undersized.  Same problem on 3D printers.  the maximum number of segments in a circle is capped.  Above a certain amount it doesn't matter, people can't tell the difference.  ironically it's the smaller circles with too many instructions that make the machine run poorly.  Gcode has G02 and G03 arc commands, but Turtle does not support arcs because it creates A LOT of work for us.

PR #573, opened 2022-02-22
Named and or instance number turtle 

PPAC37:
I try to give a name to each instantiation of the class Turtle to (update JFrame title and ) propose a file name during SaveGcode and/or SaveSVG/DXF.

This is a draft version because I have things to review and complete.
* [ ] Generators
* [ ] Converters
* [ ] Keep only the file name (without extension) and not the full path
* [ ] Propose or not to add the date at the end of the file name for the Save*.
* [ ] ...
* [ ] A cleanner way to get the version for SaveGcode

But it allowed me to notice that for a LoadImage, the Wander converter if not in CMYK, does not reset the turtle ... and adds a conversion pass. ( check once CMYK? then unchek it ( the new Wander is cumulated with the one from when CMYK? was checked)

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
+++ b/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
@@ -130,6 +130,7 @@ public final class Makelangelo {
 	private DropTarget dropTarget;
 
 	public Makelangelo() {
+		currentInstance = this;
 		logger.debug("Locale={}", Locale.getDefault().toString());
 		logger.debug("Headless={}", (GraphicsEnvironment.isHeadless()?"Y":"N"));
 		VERSION = PropertiesFileHelper.getMakelangeloVersion();
@@ -868,11 +869,17 @@ public final class Makelangelo {
         labelRangeMax.setText(Integer.toString(top));
 	}
 
+	String mainFram_pos_title = "";
 	//  For thread safety this method should be invoked from the event-dispatching thread.
 	private void createAppWindow() {
 		logger.debug("Creating GUI...");
+		if ( "true".equalsIgnoreCase(System.getenv("DEV"))){
+			mainFram_pos_title = getFullVersionString();
+		}else{
+			mainFram_pos_title = getLiteVersionString();
+		}
 
-		mainFrame = new JFrame(Translator.get("TitlePrefix") + " " + VERSION);
+		mainFrame = new JFrame(mainFram_pos_title);
 		mainFrame.setDefaultCloseOperation(JFrame.DO_NOTHING_ON_CLOSE);
 		mainFrame.addWindowListener(new WindowAdapter() {
 			@Override
@@ -1059,6 +1066,7 @@ public final class Makelangelo {
 	}
 
 	public void setTurtle(Turtle turtle) {
+		mainFrame.setTitle(turtle.getName()+" - "+mainFram_pos_title);
 		myTurtle = turtle;
 		myTurtleRenderer.setTurtle(turtle);
 		int top = turtle.history.size();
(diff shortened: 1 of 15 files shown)


src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -1091,4 +1099,20 @@ public static void main(String[] args) {
...
+	public String getFullVersionString(){
+		return getLiteVersionString()+ " ("+DETAILED_VERSION+")";		
+	}
+	public String getLiteVersionString(){
+		return Translator.get("TitlePrefix") + " " +VERSION;		
+	}
+	
+	private static Makelangelo currentInstance = null;
+	
+	public static String staticgetFullVersionString(){
```
gmembre-zenika, line 1112: the word `static` should not appear in the name of the method, we already know it is `static`.
If you have trouble with a method called in a static way and with instance, maybe you should put the code you want in a new object. Here, it could be a `Version` object wtih at least  2 static methods: `getFullVersionString()` and `getLiteVersionString()`.
You should also put this new feature in a another branch as it is not related to the naming of turtle.

coliss86, line 1112: the word static should not appear in the name of the method, we already know it is static.
If you have trouble with a method called in a static way and with instance, maybe you should put the code you want in a new object. Here, it could be a Version object wtih at least 2 static methods: getFullVersionString() and getLiteVersionString().
You should also put this new feature in a another branch as it is not related to the naming of turtle.

PPAC37: Yes you are right I mixed several stories and one of them is currently very poorly implemented.
I wanted to have your opinions if this will be retained or not (named turtle, and JFrame Title with retail version only in dev mode, and Version in gcode ) because in each case there is work and I would not want to do it if it seems useless ... / unwanted

coliss86: The version part is very interesting.
I don't know about the named turtle, what will it be used for ?

PPAC37: the idea of naming the turtles is to be able to propose a file name when you save (and to add the source file and any name + parameters values of the transformations applied as a comment) in the .gcode / . svg /.dxf output.)

A bit like with 3D printing slicers, you make it eat a "myFile".stl and it offers you when saving a "myFile"<possible other info, timestamp, ...>.gcode (and not like we currently have a file name to fill in to save ) and in the g-code you have in comment the ;MESH "name of the 3D object / name of the original file"

i-make-robots:
At the risk of making things more complicated.... there is currently no nice way to serialize to/from Turtle.  That makes it hard to `assertEquals(t0.toString(),t1.toString());` and test Turtle behavior.

i-make-robots:
Also... what problem does this solve?

i-make-robots:
Rather than name the turtle I’d open the save dialog populated with a filename generated on some stats.

PPAC37:
> Rather than name the turtle I’d open the save dialog populated with a filename generated on some stats.

Yes surely easier to implement... although... you populate your stats where and how? (That's why I thought it was necessary to do it on the turtle ... to follow it from its creation and possibly at each transformation. But it is only useful if you want to have the complete follow-up. If we just want the name of the source file, no need. )

i-make-robots:
Naming turtles is complicated because turtles can be added together and combined in exciting ways.
Saving a turtle to SVG I'd only put the creation time - what else is there?
For gcode the estimated drawing time could be added.

PPAC37:
If a turtle can be composed of turtles then this proposal (to name the turtle with the details of their transformation) becomes delicate in implementation and risks giving file names that are far too long.

I close this PR and I created ASAP one where there will be only the detailed version of Makelangelo in comment in the generated gcode and in the title bar if in dev=true mode.

I will look later to propose a file name during Save*.

PR #577, opened 2022-02-23
✅ simplify the operation of adding new tests case based of file

coliss86:
Instead of creating a test case for each file to test, now we just have to put files in the corresponding file. 
The format is : 
- fileToTest.svg
- fileToTest_expected.txt


PPAC37:
handy, thank you!

(if we delete the files (*.svg, *.sb3) ... the test will be a success?
 (possibly generate a list of files that must exist that we copy and paste into a kind of table of pre-required files? )

I think an autogenerator of _expected.txt if they don't exist would be handy as well.

coliss86:
> (if we delete the files (*.svg, *.sb3) ... the test will be a success?

Yes

> (possibly generate a list of files that must exist that we copy and paste into a kind of table of pre-required files? )

I was thinking of this possibility too. As you noticed it, I will change it.

> I think an autogenerator of _expected.txt if they don't exist would be handy as well.

This an empty file, no need to "generate" it :)

i-make-robots:
sounds good.  I'm learning a lot about unit testing from you :)

coliss86:
@PPAC37 I modified the code. What do you think of it ?

PPAC37:
It's great !
And as it goes through getRessource, "/" should remain compatible on windows. Perfect for me.

coliss86:
`/` is correct because we are looking file in the classpath not on the file system. This is the job to convert it to the correct value based on the OS.

Issue #584, opened 2022-02-25
SelectColor set a no? color when canceled


PPAC37:
**Steps to make the bug happen**
1. Go to Paper Settings
2. Click on Paper color to change the color
3. Close the color chooser or clic cancel
4. The paper color have been change to ?

**What was supposed to happen**
Have keep the old color ( cause i have cancel the color chooser )

PR #586, opened 2022-02-25
JColorChooser.showDialog return null if cancel.

PPAC37:
Close: #584


i-make-robots:
Works great!  Good catch!

PR #591, opened 2022-02-27
♻️ adjust lock as described in the javadoc

coliss86:


i-make-robots:
Not sure how this makes any functional difference.  locks before, locks after exception is handled.  please demonstrate how it can fail.

coliss86:
The log can fail after the log has been acquired and then it will be never released. It is implemented in the same way as https://github.com/MarginallyClever/Makelangelo-software/blob/master/src/main/java/com/marginallyclever/makelangelo/turtle/turtleRenderer/TurtleRenderFacade.java#L24

coliss86:
Ho, I see the contrary in the javadoc: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html, I'll update the other class

i-make-robots:
good catch.  Javadoc ftw.  As it is written in scripture by the coders of old!

PR #590, opened 2022-02-27
♻️ Extract TurtleMove constants to Enum 

coliss86:
Must be merged after #589 

This refactoring will allow adding new type of movement for #580

PR #592, opened 2022-02-27
👔 don't allow re-assigning internal value of the history of Turtle

coliss86:
To be merged after #590 
This modification prevents any one from overriding internal value of the Turtle.
The next step could be the change the visibility of history to private and create dedicated method to add / remove / replace TurtleMove. I don't know if I will implement it, I don't think it is necessary for now.


src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ChooseConnection.java
```diff
@@ -30,17 +32,22 @@
 	private static final long serialVersionUID = 4773092967249064165L;
 
 	private final ToggleButtonIcon bConnect = new ToggleButtonIcon(
-			Arrays.asList("ButtonConnect", "ButtonDisconnect"),
-			Arrays.asList("/images/connect.png", "/images/disconnect.png"),
-			Arrays.asList(Color.GREEN, Color.RED));
```
i-make-robots, line 35: this looks like it should have been one array containing two instances of a class.

src/main/java/com/marginallyclever/makelangelo/turtle/Turtle.java
```diff
@@ -281,8 +281,8 @@ private void getBounds(Point2D top,Point2D bottom) {
 	public void scale(double sx, double sy) {
 		for( TurtleMove m : history ) {
 			switch(m.type) {
-			case TurtleMove.DRAW_LINE:
-			case TurtleMove.TRAVEL:
+			case DRAW_LINE:
```
i-make-robots, line 284: why isn't this `MovementType.DRAW_LINE`?

coliss86: it is the switch on an enum, the compiler know it is on `TurtleMove`, it can't be anything else, thus it can be written like this.

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/ChooseConnection.java
coliss86: Done in https://github.com/MarginallyClever/Makelangelo-software/pull/594

PR #583, opened 2022-02-25
User custom start end gcode

PPAC37:
A try to implement #580 

* [x] (Edit : Not solve, but, worked around via the use of a JTabbedPane ) Problem with SelectPanel displaying several SelectJTextAreas (The SelectPanel and Select* seem not to want to do what I want... how to tame them?) 
* [x] implementation to use them in MarlinPlotterInferface / PlotterControl ( Thanks to coliss86 )
* [x] (will not be done. a new panel, to avoid too crowded PlotterSettings)UI review for PlotterSettings 
* [x] placeholder/replacement pattern
* [ ] Review Plotter Start/End GCode Panel ( presentation, resize, ... close on save)
~~~
G28;
G0 X{limit_left+10} Y{limit_top-20} ; Park position
M300;
~~~
give in .gcode
~~~
G28;
G0 X-315.0 Y480.0 ; Park position
M300;
~~~
* [ ] placeholder/replacement pattern cleanning
* [x] FunctionSolver.java CI test (for currently used expressions evaluation eg: +, -  )
* [ ] find the right compromise for the notion of start gcode. PlotterControl adds a G28 XY . ( so the start gcode isn't exactly one anymore... ( in fact the G28 XY should only be done if there is no G28 present in the user's start gcode... likewise during SaveGcode ) )

PR #593, opened 2022-02-27
🚧 end / start gcode

coliss86:
@PPAC37 here is where you can add the logic of start and end gcode in src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java


PPAC37:
Many thanks for all the work. You solved what was still a puzzle for me.

coliss86:
Used in #583

PR #594, opened 2022-02-28
♻️ ToggleButtonIcon: convert List to Item

coliss86:


src/main/java/com/marginallyclever/convenience/ToggleButtonIcon.java
```diff
@@ -52,4 +50,16 @@ private static ImageIcon createImageIcon(String path) {
             return null;
         }
     }
+
+    public static class Item {
```
i-make-robots, line 54: Ambiguous name is ambiguous.

coliss86: I asked myself the same but as you can only instanciate it by specifying `new ToggleButtonIcon.Item()` I think it will be OK

PR #600, opened 2022-03-04
💄 Add the filename in the title of the main window

coliss86:
When opening a file, its filename is added to the title of the main window

image

### DIFF
--- a/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
+++ b/src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
@@ -1061,7 +1061,7 @@ public final class Makelangelo {
 	private void setMainTitle(String filename) {
 		String title = "";
 		if (! "".equals(filename)) {
-			title += filename.replaceAll(".*/", "") + " - ";
+			title += new File(filename).getName() + " - ";
 		}
 		title += MakelangeloVersion.getFullOrLiteVersionStringRelativeToSysEnvDevValue();
 		mainFrame.setTitle(title);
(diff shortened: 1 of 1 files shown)


i-make-robots:
I would only display the last part of the filename.  firstly for length, and second to not reveal information in a screenshot that a user might forget about.

coliss86:
I changed the title, I also put the basename before the name of the application like in inkscape

image

i-make-robots:
Good. If we can capture that name in the right place it can be used for suggesting the save file name. The same mechanism would be used after a generator to set the generator as the suggested name.

src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -1054,6 +1058,15 @@ private void saveFile() {
 		}
 	}
 
+	private void setMainTitle(String filename) {
+		String title = "";
+		if (! "".equals(filename)) {
+			title += filename.replaceAll(".*/", "") + " - ";
```
PPAC37, line 1064: Windows Compatible '/' vs '\' ?

coliss86: Good catch, fixed

PPAC37:
Perfect when opening a file.
There remains the case of I open a file then I use a generator ... I think that the file name will not be erased or updated but I have not tested.

coliss86:
I implemented this: 
image

PPAC37:
Simple and perfect! well played!

PR #603, opened 2022-03-05
💄 Add a reset button to the settings panel

coliss86:
I also removed `totalStepperNeeded`, `totalBeltNeeded` and `totalServoNeeded` as they are not used anywhere.


i-make-robots:
The  totalStepperNeeded, totalBeltNeeded and totalServoNeeded are used by DIY people who kept asking me how much belt/cable they needed for their custom machines.

coliss86:
> The totalStepperNeeded, totalBeltNeeded and totalServoNeeded are used by DIY people who kept asking me how much belt/cable they needed for their custom machines.

ok, I'll restore them. 
I think this panel should be reorganized to emphaze this kind of information if the diy view is selected

PR #604, opened 2022-03-06
💄 Add an option to set a speed accelerator when dragging the preview

coliss86:
When click + drag on the preview pane to move the preview, I need to move the mouse a lot.

Here is a new option to configure it. By default, the value is the same as today: `1`.
image

On my computer, I set it to 4 to have a good feeling.


i-make-robots:
camera travel speed?

coliss86:
Thanks, I changed the translation

i-make-robots:
Now I have a sneaky feeling this is going to evolve into an "controls" tab where you can change the shortcuts.... lol

PR #608, opened 2022-03-10
🔊 adjust log on warning

coliss86:


src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinPlotterInterface.java
```diff
@@ -125,39 +125,39 @@ protected void onHearM114(String message) {
 
 			myPlotter.setPos(pos.x,pos.y);
 		} catch (NumberFormatException e) {
-			logger.error("M114 error: {}", message, e);
+			logger.warn("M114 error, continuing anyway: {}", message, e);
```
i-make-robots, line 128: maybe `problem` instead of `error`?

i-make-robots: i mean in the text.

Issue #602, opened 2022-03-05
Robot stuck


coliss86:
**Steps to make the bug happen**
1. draw something with the robot connected with an USB cable
2. sometimes, it get stuck and I don't understand why.

**What was supposed to happen**
The robot should recover the error and continue

**Platform (please complete the following information):**
 - Version: master
 - OS: macos
 - Robot model (if any): DYI

**Log file**
Here is the log :
```
2022-03-05 14:03:54,529 TRACE c.m.m.p.p.MarlinInterface - received 'ok' 
2022-03-05 14:03:54,530 TRACE c.m.m.p.p.MarlinInterface - sending 'N176 G1 X46.804 Y-57.418 F3000.0*126' 
2022-03-05 14:03:54,531 TRACE c.m.m.p.p.MarlinInterface - received 'ok' 
2022-03-05 14:03:54,531 TRACE c.m.m.p.p.MarlinInterface - sending 'N177 G1 X50.931 Y-62.207 F3000.0*113' 
2022-03-05 14:03:55,344 TRACE c.m.m.p.p.MarlinInterface - received 'ok' 
2022-03-05 14:03:55,344 TRACE c.m.m.p.p.MarlinInterface - received 'ok' 
2022-03-05 14:03:55,344 TRACE c.m.m.p.p.MarlinInterface - sending 'N178 G1 X55.502 Y-67.607 F3000.0*118' 
2022-03-05 14:03:55,345 TRACE c.m.m.p.p.MarlinInterface - sending 'N179 G1 X60.607 Y-73.785 F3000.0*121' 
2022-03-05 14:03:55,347 TRACE c.m.m.p.p.MarlinInterface - received 'Error:checksum mismatch, Last Line: 174' 
2022-03-05 14:03:55,347 ERROR c.m.m.p.p.MarlinInterface - Error from printer 'Error:checksum mismatch, Last Line: 174' 
2022-03-05 14:03:55,347 TRACE c.m.m.p.p.MarlinInterface - received 'Resend: 175' 
2022-03-05 14:03:55,347 TRACE c.m.m.p.p.MarlinInterface - received 'ok' 
2022-03-05 14:03:55,347 TRACE c.m.m.p.p.MarlinInterface - sending 'N175 G1 X43.051 Y-53.118 F3000.0*113' 
2022-03-05 14:03:55,349 TRACE c.m.m.p.p.MarlinInterface - received 'ok' 
[...]

PR #611, opened 2022-03-10
🩺 Improve error detection when the robot does not respond

coliss86:
Partial fix of #602 

When connecting to a robot which does not answer after 20 sec, the following error is displayed to the user : 
image

Moreover, it adds the error in the Marlin tab : 
image


src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
```diff
@@ -90,12 +90,16 @@ private void onMarlinEvent(ActionEvent e) {
...
-		case MarlinInterface.ERROR ->
-				JOptionPane.showMessageDialog(this, Translator.get("PlotterControls.Error"), Translator.get("ErrorTitle"),  JOptionPane.ERROR_MESSAGE);
+		case MarlinInterface.ERROR,
+			 MarlinInterface.DID_NOT_FIND,
+			 MarlinInterface.COMMUNICATION_FAILURE -> {
+			if (!isErrorAlreadyDisplayed) {
+				JOptionPane.showMessageDialog(this, Translator.get("PlotterControls." + e.getActionCommand()), Translator.get("ErrorTitle"), JOptionPane.ERROR_MESSAGE);
+				isErrorAlreadyDisplayed = true;
+			}
+		}
```
i-make-robots, line 100: test here that MarlinPlotterInterface remembers the line number (A) so that after reconnect and home it can resume from line (A-1).  but how to automate such a test?

i-make-robots (approved):

src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/PlotterControls.java
coliss86: hum quite challenging, I'm opening an issue with this idea 💪

coliss86:
@i-make-robots ready for review :)

src/main/java/com/marginallyclever/makelangelo/plotter/plottercontrols/MarlinInterface.java
```diff
@@ -103,14 +105,21 @@ private void onClose() {
 	}
 	
 	private void onTimeoutCheck() {
-		if (System.currentTimeMillis() - lastReceivedTime > TIMEOUT_DELAY) {
-			logger.trace("Heartbeat: M400");
-			chatInterface.sendCommand("M400");
+		long delay = System.currentTimeMillis() - lastReceivedTime;
+		if (delay > TIMEOUT_DELAY) {
+			if (delay > TIMEOUT_DELAY * 5) {
```
i-make-robots, line 110: this should probably be a constant like `TIMEOUT_DELAY`

src/main/resources/languages/english.xml
```diff
@@ -736,6 +736,11 @@
 		<value>Error in communicating with the robot.</value>
 		<hint>text</hint>
 	</string>
+	<string>
+		<key>PlotterControls.CommunicationFailure</key>
+		<value>No answer from the robot, please check the USB cable, speed settings...</value>
```
i-make-robots, line 741: what does "..." mean?  as a novice user this would be confusing.

coliss86: what do you propose ? I can remove the three dots for clarity if you want.

src/main/java/com/marginallyclever/makelangelo/plotter/plottercontrols/MarlinInterface.java
coliss86: I have updated the code

i-make-robots:
If the communication delay persists, does the message keep reappearing?

src/main/resources/languages/english.xml
i-make-robots: I don't immediately have a good suggestion.  Let's talk about it, maybe see what users would like.

PR #619, opened 2022-03-17
👷 change the CI to ubuntu to check file case sensitiveness

coliss86:


coliss86:
why did you add those commits ? everything is done in https://github.com/MarginallyClever/Makelangelo-software/pull/618.

Also, the jar nodegraph is taken from jitpack now, you don't need to add it in the repos.

coliss86:
I will reopen another one with ubuntu only

Issue #625, opened 2022-03-22
Save a portion of the turtle


i-make-robots:
**Is your feature request related to a problem? Please describe.**
User's pen died part way into a many-days drawing.

**Describe the solution you'd like**
Move the double-slider at the bottom of the main window to isolate a portion of the vector.
save as > any format

Only the visible portion of the Turtle should be exported.  Should respect color changes, etc.

i-make-robots:
Ugh.  There are two paths to save files: 
- File > save / save as...
- Makleangelo > Save GCode

The only way I see to make it work is to deliver a turtle *already trimmed* to the save system.

i-make-robots:
Correction!  GCode is not an option from file > save....
Huh.

PR #631, opened 2022-03-25
save only the portion of the drawing shown by the range slider. (#625)

i-make-robots:
for #625

Tested on stanford bunny, seems fine.
Tested on a jpeg.  Correct gcode, but loading gcode into the app smashes all colors to black.
Turtles might want to generate a comment after M0 with the hex value of the color.


src/main/java/com/marginallyclever/makelangelo/Makelangelo.java
```diff
@@ -876,6 +877,8 @@ private void onChangeSlider(ChangeEvent e) {
         myTurtleRenderer.setLast(top);
         labelRangeMin.setText(Integer.toString(bottom));
         labelRangeMax.setText(Integer.toString(top));
+		SaveGCode.setTrimHead(bottom);
```
coliss86, line 880: Settings static values in the a bad pattern. Every time the user ask to save the gcode, a new `SaveGCode` is instanciated, so thoses values can be set in the constructor of even better in the `save()` method

coliss86 (changes requested):
Remove the static values and add units tests please

i-make-robots:
Show me an elegant way to remove the static variable. I did want it, either.

coliss86:
Here is : https://github.com/MarginallyClever/Makelangelo-software/pull/632

coliss86 (approved):
LGTM

PR #635, opened 2022-03-29
LoadFilePanel a JTextArea for long file path and name

### LINKED ISSUE #634: Very long absolute path in Load image dialog
**Steps to make the bug happen**
1. Load an image with a very long absolute path and filename.

**What was supposed to happen**
probably truncate the filename in the dialog somehow so it doesn't resize in a silly way.

**Platform (please complete the following information):**
 - Version: 7.23.0
 - OS: Windows

PPAC37:
close #634 

I would prefer an uneditable JTextArea with line warping than a JLabel with trim or "...".
But (unless we revise the layout) it makes a big button if on multiple lines...

image


i-make-robots:
I would rather trim the /home end of the line than the .png end of the line.  what do you think?

PR #642, opened 2022-04-01
Better reordering

i-make-robots:


coliss86 (approved):
nice !
how does it work ?

i-make-robots:
1. Split turtles into segments (pair of points representing a draw move)
2. Do a greedy tour and collect adjacent segments into linecollections.
3. Run it again looking for line collections that touch each other. (New)
4. Put everything back together.

It does not look for the shortest travel moves.

PR #644, opened 2022-04-05
added flow field generator and converter

i-make-robots:


src/main/resources/languages/english.xml
```diff
@@ -2194,4 +2194,37 @@
 		<key>Converter_EdgeDetection.sampleSize</key>
 		<value>SampleSize</value>
 	</string>
+
+	<string>
+		<key>Generator_FlowField.name</key>
+		<value>Flow field</value>
+	</string>
+	<string>
+		<key>Generator_FlowField.name</key>
```
PPAC37, line 2203: duplicate (to be deleted to pass the CI)
Note to myself : set as trace the expected errors (test of the test) so the real errors become obvious in the logs of the buid...

i-make-robots: yes please.  every time I am hunting for the value in the extensive error report.

Issue #492, opened 2022-01-29
svg not rendered completely


coliss86:
**Steps to make the bug happen**
1. Load the attached svg
2. 2 parts are missings: "Joyeux Anniversaire" and the robot on the lower left
![image](https://user-images.githubusercontent.com/23615562/151664860-2b9ddb71-bf20-4beb-b5b2-3920f0ba5b5f.png)

This file has been created with inkscape: [original file](https://user-images.githubusercontent.com/23615562/151664869-0f364b5e-fec0-4f28-bee9-55de4ac5b4c8.svg)

**What was supposed to happen**
the svg looks like this:
![image](https://user-images.githubusercontent.com/23615562/151664890-7527fa1c-ed7d-4619-83dd-0ce65d769a7f.png)

**Platform (please complete the following information):**
 - Version: master
 - OS: OSX
 - Robot model (if any): [e.g. Makelangelo 5]

PPAC37:
it seems that the text must be transformed into paths before saving in svg under inkscape (do a Shift+Ctrl+C on the text) but even with that, the svg obtained does not look better on Makelangelo...

there is surely the fact that the batik lib is tu update ... and/or the fact that the system fonts are not necessarily well compatible with the JRE (when we do not transform the text on my way ) )

a workaround when the SVG does not render as you want is an external transformation of an SVG format into EPS with inkscape, then EPS into DXF format with pstoedit ... cf : https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Other_2D_formats 

Of course it's not the easiest but it works well (especially when you have everything installed inkscape and pstoedit and make to use the Makefile 

Create a directory put the .svg and the makefile and do make from a console. (Makefile in the zip with the svg and the dxf convertion output)
[svg_to_eps_to_dxf_via_make.zip](https://github.com/MarginallyClever/Makelangelo-software/files/7964240/svg_to_eps_to_dxf_via_make.zip)

```
$ make
inkscape -E joyeux_anniv.eps joyeux_anniv.svg
Gtk-Message: 18:41:19.073: Failed to load module "canberra-gtk-module"
pstoedit -dt -f "dxf: -polyaslines -mm" joyeux_anniv.eps joyeux_anniv.dxf
pstoedit: version 3.70 / DLL interface 108 (built: Aug 29 2017 - release build - g++ 7.2.0 - 64-bit) : Copyright (C) 1993 - 2014 Wolfgang Glunz
[...]

PR #645, opened 2022-04-05
add moditect

coliss86:
The application starts !
but not in IntelliJ, only in the console with `./mvnw clean install` and then `java -jar target/Makelangelo-7.32.0-with-dependencies.jar`.
image

To be able to run it directly from IntelliJ, the xalan jar must be patched :'( This is why I always remove the `module-info.java` from my working copy.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -332,7 +332,7 @@ It pairs really well with Marlin-polargraph, the code in the brain of the robot 
 										<version>${xalan.version}</version>
 									</artifact>
 									<moduleInfoSource>
-										module xalan.xalan {
+										module xalan {
 										}
 									</moduleInfoSource>
 								</module>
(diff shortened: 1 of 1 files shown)


i-make-robots:
does it run problem SVG files better than before?

i-make-robots:
Is there some generated module-info.java file that we can extract from the final jar and include earlier in the build to help intellij?  maybe bump up when the moditect runs?

coliss86:
> does it run problem SVG files better than before?

Nop, the problem raised in #492 is still present: 
image

coliss86:
So maybe there is a bug in makelangelo :'( I really don't know where to dig...

i-make-robots:
> only in the console with ./mvnw clean install and then java -jar target/Makelangelo-7.32.0-with-dependencies.jar.
I'm still not able to run test, compile, package, or install.  They all require test, which fails at boot layer because of BSFManager.

coliss86:
yep :'(

i-make-robots:
So apparently the maintainer of EchoSVG is going to remove the need for Xalan to simplify everyone's life.  Thank you to that maintainer!  Woo!

coliss86:
waou nice !

PR #648, opened 2022-04-09
fix voronoi diagrams and related converter

i-make-robots:
Starting to look good

- [x] in a "one time" converter all changes immediately restart the conversion process.  in iterative converters some options should *not* restart the entire process.
- [x] iterative methods should wipe the master Turtle when they start or artifacts will show through.
- [x] when "pause" is on and the dialog is still open generate the final output (assume it is done).
- [x] changing the source image appears to make the GUI unresponsive.  fix!
- [x] voronoi cells with center.y<0 appear to drift over time towards y=0, reason unknown.
- [x] sometimes a second thread starts and the first one never ends.
- [x] iterative methods should have their own "pause" button and maybe a visible iteration count
- [ ] iterative methods have no way to show that they're still going, tho they may be very slow for large data sets.


src/test/java/com/marginallyclever/convenience/TestConvexHull.java
```diff
@@ -0,0 +1,79 @@
...
+        hull.add(new Vector2d(17, 5));
+        hull.add(new Vector2d(-3, 15));
+        hull.add(new Vector2d(-3, -9));
+        hull.add(new Vector2d(0, 11));
+        hull.add(new Vector2d(-9, -3));
+        hull.add(new Vector2d(-4, -2));
+        hull.add(new Vector2d(12, 10));
+        assertEquals("ConvexHull{hull=[(-9.0, -3.0), (-3.0, -9.0), (19.0, -8.0), (17.0, 5.0), (12.0, 17.0), (5.0, 19.0), (-3.0, 15.0)]}", hull.toString());
+        Vector2d center = hull.getCenterOfHull();
+        assert(hull.contains(center));
```
coliss86, line 41: `assert()` is a method from the jvm which will fail only if you launch it with the `-ea` parameter. Here, you must use `assertTrue()`

coliss86:
Nice work, don't forget to update the title of the PR to have a good commit message when merging

src/test/java/com/marginallyclever/convenience/TestConvexHull.java
i-make-robots: Good catch!

Issue #650, opened 2022-04-16
Save Turtle output to JPEG


i-make-robots:
**Is your feature request related to a problem? Please describe.**
User Mesut asked if there was a way to save out the image on the virtual paper to a bitmap.
The default bitmap size would be 1 pixel = 1mm of paper.

**Describe the solution you'd like**
There is already a `com.marginallyclever.makelangelo.donatelloimpl.nodes.TurtleToBufferedImage`.  The core feature could be reused to generate the bitmap, then exported with `ImageIO` to the destination of choice.

**Describe alternatives you've considered**
It's not possible to get a Graphics context that works with the GLJPanel, so capturing screenshots is not possible using this technique.

**Additional context**
It would be nice if the user could then raise the number of pixels per mm, but it gets complicated fast.

i-make-robots:
@PPAC37 thank you for proving me wrong! 🥇 With screenshots we can do more automated testing, I hope.  Requires very careful changes to the look + feel...

i-make-robots:
Addressed in  #654

PR #651, opened 2022-04-16
take a screen shot ( poc draft )

PPAC37:
A test capture of a component and screenshot of this one. 
To check if it works or not with other OS. 
Draft to adapt and clean ... or to use as an example to achieve a cleaner implementation.
for #650 (but actualy save as png )


i-make-robots:
Works on windows 11.  Takes three screenshots.

PPAC37:
this PR is no longer relevant. because a method of saving to bitmap exists since #654

PR #654, opened 2022-04-26
Save vector in app to bitmap

i-make-robots:
Wow, this was a can of worms.  saving to bitmap needed a new file comparison tool (binary to binary), then SVG had issues, which caused a test to change, which affected all other save formats.  gross!

Also: having saveGCode be the only save not in the same package is UGH.


i-make-robots:
Also it looks like my unit tests are *never* being deleted from the windows temp folder.  wtf?

src/main/java/com/marginallyclever/makelangelo/plotter/plottercontrols/MarlinPlotterInterface.java
```diff
@@ -203,8 +203,23 @@ private static String getColorName(int toolNumber) {
...
+			default -> "0x" + paddedHex(toolNumber); // display unknown RGB value as hex
 		};
 		return name;
 	}
+
+	/**
+	 * @param number the number to convert to hex
+	 * @return the hex number returned is 6 characters long.  May pad the left end with zeros.
+	 */
+	private static String paddedHex(int number) {
```
coliss86, line 215: Can you please move this method to `com.marginallyclever.convenience.StringHelper` and add some unit tests to it ?

coliss86 (changes requested):
LGTM ! nice work !

Issue #587, opened 2022-02-26
a color value association to color name does not match w3c color name for green...


PPAC37:
if we want to name the color as in https://www.w3.org/TR/css-color-3/#svg-color

https://github.com/MarginallyClever/Makelangelo-software/blob/33bc26eedbb22490785d33ac550ab710d3fb1a56/src/main/java/com/marginallyclever/makelangelo/plotter/plotterControls/MarlinPlotterInterface.java#L193-L201

~~~
		case 0x00ff00:// = w3c "lime" https://www.w3.org/TR/css-color-3/#normative
			name = "lime";
			break;
		case 0x008000:// = w3c "green" https://www.w3.org/TR/css-color-3/#normative
			name = "green";
			break;
~~~

yes it's not a big deal but ...

i-make-robots:
It would be a good idea to have a matrix? Tree? of styles and colors. A style might be “comic” and then the sun elements are colic marker designations and their hex code. Another would be pantone. Etc.

PPAC37:
You think of color sets that match some theme ?
like "christmas" = { "red", "green", "white", ... }
and 
![image](https://user-images.githubusercontent.com/94939582/155855591-fc6cdd44-3da6-4fc0-88ae-d1f0031753e8.png)

Or a way to present / group the colors that would be close to each other in a color chart ?
( a bit like what is done on https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Transformations#color )

i-make-robots:
No... I was thinking more like official color codes from pen makers.  [copic](https://copic.jp/en/about/color-system/), staedler (can't find), and (pantone)[https://www.pantone-colours.com/].

PPAC37:
maybe there is a way around this, like making a voluntary typo so that it is not exactly the same name but that it is easily guessed like Bleu => BeBleu , Red => ReRed? ;)

coliss86:
why not just the name as defined in the w3c ? no copyright problem and names on the list are already chosen...

coliss86:
like this: https://www.w3.org/wiki/CSS/Properties/color/keywords

PPAC37:
Yes to use W3C color naming.
It will just be necessary to agree on when there are several names for the same value which one is used.

ex : 
~~~

 case 0x00ffff: name = "aqua"; // [aqua, cyan]
break;

 case 0xff00ff: name = "fuchsia"; // [fuchsia, magenta]
break;

[...]

PR #655, opened 2022-04-27
use W3C color names in MarlinPlotterInterface (for #587)

i-make-robots:


src/main/java/com/marginallyclever/convenience/W3CColorNames.java
```diff
@@ -164,4 +169,16 @@ public class SVGColorNames {
     public static ColorRGB get(String name) {
         return map.get(name);
     }
+
+    /**
+     * @param match the color to matcch
+     * @return the name matching the given color, or null if no match.
+     */
+    public static String get(ColorRGB match) {
+        for(String key: map.keySet()) {
```
coliss86, line 178: you don't have to iterate through the map, you can use : `map.get(match)`

coliss86: And the method `isEqualsTo()` can be removed and the method `equals()` can be used in place.

i-make-robots: There is only map.get(key), no map.get(value).  I am doing the latter.

isEqualsTo() is a weird method.  removed.  ty!

coliss86, line 178: Have you considerd doing it with `stream` ? Here is an example : https://www.baeldung.com/java-map-key-from-value#a-functional-approach

i-make-robots: Am I correct that stream would go through all items in the map?  My way it stops on the first hit.  slightly faster.

coliss86: no, stream won't go to the end of the iteration, this is why they are built for.

PR #660, opened 2022-07-14
remove the filter for serial connections on mac to return all the values possible. Useful for socat.

coliss86:
Asked on discord by @Mieg : 
> I tried setting up a 'virtual' serial device on my macbook using socat, and that does work - but Makelangelo doesn't see it because the device name doesn't start with /dev/tty. or /dev/cu.


src/main/java/com/marginallyclever/communications/serial/SerialTransportLayer.java
```diff
@@ -22,6 +22,7 @@
 public class SerialTransportLayer implements TransportLayer {
 
 	private static final Logger logger = LoggerFactory.getLogger(SerialTransportLayer.class);
+	public static final String CU_USBSERIAL = "cu.usbserial";
```
i-make-robots, line 25: nice.

PR #673, opened 2022-11-23
fix #672

### LINKED ISSUE #672: Add huge to plotter settings
**Is your feature request related to a problem? Please describe.**
M3.3, M5, Huge... there may be other plotters in future. 
It is currently very hard to change the machine seen and calculated to be anything other than an M5.

**Describe the solution you'd like**
A drop-down list of available presets would be great.  
Adding new presets by tweaking an existing model and hitting save would be good, too.

**Describe alternatives you've considered**
See Prusa Slic3r advanced settings tab for examples.
image

**Additional context**

i-make-robots:
It was previously very difficult to manage the various records of plotter settings (aka "profiles").  
a profileManager and an profileManagerPanel are introduced.  the panel includes a selection tool, a copy-and-rename profile tool, and a delete profile tool.
the plotter settings and plotter custom gcode panels are then added to this new profile manager panel to better logically consolidate settings.
In a future update plotter render style might also be moved here.


i-make-robots:
Does not appear to delete profiles from backing storage.  When I click - the item disappears from the comboBox...until the next time the settings are opened, and then it's back.

PR #678, opened 2023-02-05
jogamp updated to version 2.4.0

jsikorsky:
Previous version 2.4.0-rc-20210111 of jogamp did not work on MacOS with Apple Silicon chip since it included only x86_64 libraries.
Version 2.4.0 is multiplatform.
The build runs without problems on macos arm64 (Macbook M1), did not test other platforms.

Please note, src/main/assembly/application.xml contains errors in excludes and includes. This causes that include and exclude statements do not work at all, but this makes no harm as well. The file was left unchanged.


i-make-robots:
seems to run in windows...

i-make-robots:
@jsikorsky great PR!  Any more ideas what to do about src/main/assembly/application.xml ?

jsikorsky:
You can just leave at is, because excludes and includes are incorrect (these are relative paths in jar, not project coordinates) and Maven warns about this.
Or you can leave it off altogether - please see the commit above. 
I'm not a Java and Maven person, had to learn this all yesterday, but I'm quite sure it will have no other consequences.

jsikorsky:
I mean, let off just includes and excludes, unpack=true is important there.
Last commit works fine on Apple arm64.

PR #699, opened 2023-09-04
Intensity to height converter with better artistic style

itsMohammedThaier:
- `waveIntensity` maximum amplitude for a wave.
- Got rid of `maxHeight`

Old style: max wave amplitude is `maxHeight`, and 'z_scale' is determined by average color of a rect area that's directly relative to `maxHeight`:
Before

New style: max wave amplitude is `waveIntensity`, and 'z_scale' is determined by average color of a line of pixels:
After

Although it's a mere artistic choice, I prefer my style since it can capture details better(notice in the second-second image how you can still see the face even with high "Wave Spacing" value).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
 	<groupId>com.marginallyclever</groupId>
 	<artifactId>Makelangelo</artifactId>
-	<version>7.40.3</version>
+	<version>7.41.0</version>
 	<name>Makelangelo</name>
 		<description>Makelangelo Software is a Java program that prepares art for CNC plotters.  It is especially designed for the Makelangelo Robot.
 It pairs really well with Marlin-polargraph, the code in the brain of the robot that receives instructions and moves the motors.</description>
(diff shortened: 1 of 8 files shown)


i-make-robots:
This is a well done PR.  please update from the parent branch.

i-make-robots:
so you changed the name of maxHeight to waveIntensity?  I'm still confused what this does.  I don't seem to be able to get this branch to test what you've done.  I'm doing something wrong and I feel a little dumb...

itsMohammedThaier:
> so you changed the name of maxHeight to waveIntensity? I'm still confused what this does. I don't seem to be able to get this branch to test what you've done. I'm doing something wrong and I feel a little dumb...


- No; `waveIntensity` is much different than `maxHeight `although they're meant to partially represent a similar concept which is the maximum amplitude of a wave.

- Got rid of `maxHeight` and all its relatives(such as its slider and getters\setters, translator reference..etc).

- Explaining the artistic difference would be much easier in a call so let me know when you're up for a discord call.

PR #712, opened 2023-10-18
Weighted lines

i-make-robots:
Take any arbitrary set of lines and make them thicker based on a source image.


i-make-robots:
Also created DirectionLoopTurtleRenderer which draws each loop starting from red and blending to blue.

PR #746, opened 2024-05-27
Updated german translation

peter-pri:
Hello Dan,
I updated the most obvious missing German translations.


i-make-robots:
awesome, thank you!

PR #747, opened 2024-05-28
Update german, some additional translations. This is nearly the remaining part.

peter-pri:


i-make-robots:
I would love to come up with a strategy that would make it easier for you to complete translations.  Any ideas?

peter-pri:
My strategies so far have been:
1. Check every menu entry in the user interface manually if a German text appears or the the symbolic key name.
2. Check  the results of automated tests, which translations are missing.

Results until now: All translations tests are passed.
One translation is missing in my last change:
	<string>
        <key>Converter_CMYK_Crosshatch.Note</key>
        <value><![CDATA[Zeichnet Gelb, Cyan, Magenta, und Schwarz.]]></value>
        <hint>[<![CDATA[Draws Yellow, Cyan, Magenta, and then Black.]]>]</hint>
    	</string>
Initially I had a typo in there, so I removed it to get it working.

It would be helpful if the english.xml file has the same formatting as the other languages.
This can be easily done e.g. with the vscode extension XML Tools form Josh Johnson.
In addition the file content should be sorted alphabetically by the keys to make it comparable with a Linux tool like meld.
With this sorting task I was not successful until now.

On other idea to handle missing translations, is to show instead of the key, the English translation in red.
So the user can make use of the existing translations and see what is missing.

i-make-robots:
If anything I wish the other language files were more like the english one.  Scrolling forever is painful.
I completely agree that sorting by key should be a thing.
I would prefer to show the key name instead of the english value.

peter-pri:
OK, I see.
I translate some additional entries.
I don' t know why I didn't see them yesterday.

